PR #15, opened 2016-07-22
Added an implementation of Event System

ritave:
This module allows for listening to interesting Gateway events and thus allowing development of Gateway plugins.

### DIFF
--- a/core/inc/gateway_ll.h
+++ b/core/inc/gateway_ll.h
@@ -49,6 +49,21 @@ typedef struct GATEWAY_PROPERTIES_DATA_TAG
 	VECTOR_HANDLE gateway_properties_entries;
 } GATEWAY_PROPERTIES;
 
+/** @brief  Enum representing different gateway events that have support for callbacks. */
+typedef enum GATEWAY_EVENT_TAG
+{
+	GATEWAY_CREATED = 0,
+	GATEWAY_DESTROYED,
+	/* @brief  Not an actual event, used to keep track of count of different events */
+	GATEWAY_EVENTS_COUNT
+} GATEWAY_EVENT;
+
+/** @brief  Placeholder for possible event context in the future. Right now NULL will be passed. */
+typedef void* GATEWAY_EVENT_CTX;
+
+/** @brief	Function pointer that can be registered and will be called for gateway events */
+typedef void(*GATEWAY_CALLBACK)(GATEWAY_HANDLE gateway, GATEWAY_EVENT event_type, GATEWAY_EVENT_CTX context);
+
 /** @brief		Creates a new gateway using the provided #GATEWAY_PROPERTIES.
 *
 *	@param		properties		#GATEWAY_PROPERTIES structure containing 
@@ -83,6 +98,14 @@ extern MODULE_HANDLE Gateway_LL_AddModule(GATEWAY_HANDLE gw, const GATEWAY_PROPE
 *	@param		module	Pointer to a #MODULE_HANDLE that needs to be removed.
 */
 extern void Gateway_LL_RemoveModule(GATEWAY_HANDLE gw, MODULE_HANDLE module);
+
+/** @brief  Registers a function to be called on a callback thread when #GATEWAY_EVENT happens
+*
+*   @param  gw         Pointer to a #GATEWAY_HANDLE to which register callback to
+*   @param  event_type Enum stating on which event should the callback be called
+*   @param  callback   Pointer to a function that will be called when the event happens
+*/
+extern void Gateway_AddEventCallback(GATEWAY_HANDLE gw, GATEWAY_EVENT event_type, GATEWAY_CALLBACK callback);
 
 #ifdef UWP_BINDING
 
--- a/core/inc/internal/event_system.h
+++ b/core/inc/internal/event_system.h
@@ -1,0 +1,30 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+/** @file       event_system.h
+*   @brief      Header file with internal API for managing the callback system
+*/
+
+#ifndef EVENT_SYSTEM_H
+#define EVENT_SYSTEM_H
+
+#include "gateway_ll.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif // __cplusplus
+
+typedef struct EVENTSYSTEM_DATA* EVENTSYSTEM_HANDLE;
+
+extern EVENTSYSTEM_HANDLE EventSystem_Init(void);
+extern void EventSystem_AddEventCallback(EVENTSYSTEM_HANDLE event_system, GATEWAY_EVENT event_type, GATEWAY_CALLBACK callback);
+extern void EventSystem_ReportEvent(EVENTSYSTEM_HANDLE event_system, GATEWAY_HANDLE gw, GATEWAY_EVENT event_type);
+extern void EventSystem_Destroy(EVENTSYSTEM_HANDLE event_system);
+
+#ifdef __cplusplus
+}
+#endif // __cplusplus
+
+
+#endif // !EVENT_SYSTEM_H
--- a/core/src/gateway_ll.c
+++ b/core/src/gateway_ll.c
@@ -13,6 +13,7 @@
 #include "gateway_ll.h"
 #include "message_bus.h"
 #include "module_loader.h"
+#include "internal/event_system.h"
 
 typedef struct GATEWAY_HANDLE_DATA_TAG {
 	/** @brief Vector of MODULE_DATA modules that the Gateway must track */
@@ -20,6 +21,9 @@ typedef struct GATEWAY_HANDLE_DATA_TAG {
 
 	/** @brief The message bus contained within this Gateway */
 	MESSAGE_BUS_HANDLE bus;
+
+	/** @brief Handle for callback event system coupled with this Gateway */
+	EVENTSYSTEM_HANDLE event_system;
 } GATEWAY_HANDLE_DATA;
 
 typedef struct MODULE_DATA_TAG {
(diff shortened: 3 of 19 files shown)


core/inc/gateway_ll.h
```diff
@@ -49,6 +49,21 @@ typedef struct GATEWAY_PROPERTIES_DATA_TAG
...
+	GATEWAY_DESTROYED,
+	/* @brief  Not an actual event, used to keep track of count of different events */
+	GATEWAY_EVENTS_COUNT
+} GATEWAY_EVENT;
+
+/** @brief  Placeholder for possible event context in the future. Right now NULL will be passed. */
+typedef void* GATEWAY_EVENT_CTX;
+
+/** @brief	Function pointer that can be registered and will be called for gateway events */
+typedef void(*GATEWAY_CALLBACK)(GATEWAY_HANDLE, GATEWAY_EVENT, GATEWAY_EVENT_CTX context);
```
avranju: Maybe add parameter names for the first 2 parameters as well?

core/inc/gateway_ll.h
```diff
@@ -84,6 +99,14 @@ extern MODULE_HANDLE Gateway_LL_AddModule(GATEWAY_HANDLE gw, const GATEWAY_PROPE
 */
 extern void Gateway_LL_RemoveModule(GATEWAY_HANDLE gw, MODULE_HANDLE module);
 
+/** @brief  Registers a function to be called on a callback thread when #GATEWAY_EVENT happens
+*
+*   @param  gw         Pointer to a #GATEWAY_HANDLE to which register callback to
+*   @param  event_type Enum stating on which event should the callback be called
+*   @param  callback   Pointer to a function that will be called when the event happens
+*/
+extern void Gateway_AddEventCallback(GATEWAY_HANDLE gw, GATEWAY_EVENT event_type, GATEWAY_CALLBACK callback);
```
avranju: Interesting that this returns `void`. Can this not fail?

core/inc/internal/event_system.h
```diff
@@ -0,0 +1,30 @@
...
+#ifdef __cplusplus
+extern "C"
+{
+#endif // __cplusplus
+
+typedef struct EVENTSYSTEM_DATA* EVENTSYSTEM_HANDLE;
+
+extern EVENTSYSTEM_HANDLE EventSystem_Init(void);
+extern void EventSystem_AddEventCallback(EVENTSYSTEM_HANDLE event_system, GATEWAY_EVENT event_type, GATEWAY_CALLBACK callback);
+extern void EventSystem_ReportEvent(EVENTSYSTEM_HANDLE event_system, GATEWAY_HANDLE gw, GATEWAY_EVENT event_type);
```
avranju: Can `EventSystem_AddEventCallback` and `EventSystem_ReportEvent` fail? If yes, should they return a status code of some kind?

core/src/gateway_ll.c
```diff
@@ -85,19 +93,30 @@ GATEWAY_HANDLE Gateway_LL_Create(const GATEWAY_PROPERTIES* properties)
...
+					if (gateway->event_system == NULL)
+					{
+						LogError("Gateway_LL_Create(): Unable to initialize callback system");
+						gateway_destroy_internal(gateway);
+						gateway = NULL;
+					}
+					else
+					{
+						/*Codes_SRS_GATEWAY_LL_26_001: [ This function shall initialize attached Gateway Events callback system and report GATEWAY_CREATED event. ] */
+						EventSystem_ReportEvent(gateway->event_system, gateway, GATEWAY_CREATED);
```
avranju: Since the gateway just got created, will there be any listeners at this point, i.e., will anyone ever get notified of `GATEWAY_CREATED`?

core/src/internal/event_system.c
```diff
@@ -0,0 +1,333 @@
...
+#include "azure_c_shared_utility/condition.h"
+#include "azure_c_shared_utility/list.h"
+
+#include "gateway_ll.h"
+#include "internal/event_system.h"
+
+#include <assert.h>
+#include <stdlib.h>
+
+struct EVENTSYSTEM_DATA {
```
avranju: Do we want the `typedef struct` thing here too (like `THREAD_QUEUE_ROW` below)?

core/src/internal/event_system.c
```diff
@@ -0,0 +1,333 @@
...
+
+void EventSystem_AddEventCallback(EVENTSYSTEM_HANDLE event_system, GATEWAY_EVENT event_type, GATEWAY_CALLBACK callback)
+{
+	/* Codes_SRS_EVENTSYSTEM_26_012: [ This function shall log a failure and do nothing else when either `event_system` or `callback` parameters are NULL. ] */
+	if (event_system == NULL || callback == NULL)
+	{
+		LogError("null gateway events handle or callback parameters during event registration");
+	}
+	/* Codes_SRS_EVENTSYSTEM_26_011: [ This function shall register given GATEWAY_CALLBACK and call it when given GATEWAY_EVENT event happens inside of the gateway. ] */
+	else if (VECTOR_push_back(event_system->event_callbacks[event_type], &callback, 1) != 0)
```
avranju: Should this be inside a lock?

core/src/internal/event_system.c
```diff
@@ -0,0 +1,333 @@
...
+		int real_is_errored = 0;
+		Lock(event_system->internal_change_lock);
+		real_is_errored = event_system->is_errored;
+		Unlock(event_system->internal_change_lock);
+		
+		if (!real_is_errored)
+		{
+			/* We need to copy the callback queue because the callback might register another function */
+			/* Codes_SRS_EVENTSYSTEM_26_007: [ This function shan't call any callbacks registered for any other GATEWAY_EVENT other than the one given as parameter. ] */
+			VECTOR_HANDLE callbacks = event_system->event_callbacks[event_type];
```
avranju: Should this read be under a lock?

core/src/internal/event_system.c
```diff
@@ -0,0 +1,333 @@
...
+					/*Codes_SRS_EVENTSYSTEM_26_013: [ Should the worker thread ever fail to be created or any internall callbacks fail, failure will be logged and no further callbacks will be called during gateway's lifecycle. ] */
+					LogError("Failed to create call queue during event report");
+					event_system->is_errored = 1;
+				}
+				else
+				{
+					if (VECTOR_push_back(call_queue, VECTOR_front(callbacks), vector_size) != 0)
+					{
+						/*Codes_SRS_EVENTSYSTEM_26_013: [ Should the worker thread ever fail to be created or any internall callbacks fail, failure will be logged and no further callbacks will be called during gateway's lifecycle. ] */
+						LogError("Failed to copy callback queue during event report");
```
avranju: Should the `call_queue` vector get freed here?

core/src/internal/event_system.c
```diff
@@ -0,0 +1,333 @@
...
+				VECTOR_HANDLE call_queue = VECTOR_create(sizeof(GATEWAY_CALLBACK));
+				if (call_queue == NULL)
+				{
+					/*Codes_SRS_EVENTSYSTEM_26_013: [ Should the worker thread ever fail to be created or any internall callbacks fail, failure will be logged and no further callbacks will be called during gateway's lifecycle. ] */
+					LogError("Failed to create call queue during event report");
+					event_system->is_errored = 1;
+				}
+				else
+				{
+					if (VECTOR_push_back(call_queue, VECTOR_front(callbacks), vector_size) != 0)
```
avranju: It looks like you're copying the callback pointers into a temporary vector to essentially take a snapshot of the registered listeners so that we can safely iterate through the vector without needing to lock each access into the vector. If that's the intent here then we should probably do this copy inside a lock.

core/src/internal/event_system.c
```diff
@@ -0,0 +1,333 @@
...
+		}
+		list_destroy(handle->thread_queue);
+
+		for (int i = 0; i < GATEWAY_EVENTS_COUNT; i++)
+			VECTOR_destroy(handle->event_callbacks[i]);
+		free(handle);
+	}
+}
+
+static void callbacks_call(EVENTSYSTEM_HANDLE event_systen, GATEWAY_HANDLE gw, GATEWAY_EVENT event_type, VECTOR_HANDLE callbacks)
```
avranju: This parameter should probably be called `event_system`.

core/src/internal/event_system.c
```diff
@@ -0,0 +1,333 @@
...
+		}
+	}
+
+	Lock(event_system->internal_change_lock);
+
+	event_system->callback_thread = NULL;
+
+	Unlock(event_system->internal_change_lock);
+
+	ThreadAPI_Exit(THREADAPI_OK);
```
avranju: Doesn't calling `ThreadAPI_Exit` mean that the next line doesn't get to run?

samples/callbacks_sample/src/main.c
```diff
@@ -0,0 +1,26 @@
...
+
+int main(int argc, char** argv)
+{
+    GATEWAY_HANDLE gw = Gateway_LL_Create(NULL);
+    Gateway_AddEventCallback(gw, GATEWAY_DESTROYED, my_callback);
+	Gateway_AddEventCallback(gw, GATEWAY_DESTROYED, second_callback);
+    printf("Hello with running gateway!\n");
+    Gateway_LL_Destroy(gw);
+	printf("Gateway was destroyed\n");
+}
```
avranju: This sample needs to be cleaned up a bit. I'd expect it to showcase handling all event types and print meaningful messages to the console (for e.g. the event handler for `GATEWAY_CREATED` might print: *The gateway was created.*).

core/inc/gateway_ll.h
ritave: @avranju Yes, it can fail. But since the whole event system is multi-threaded and to check for all the errors would need either another callback or error polling I've decided for the simplest way possible for now - disable all callbacks and log an error (see requirements doc), until there's a need for better error handling.

core/inc/internal/event_system.h
ritave: @avranju See comment for `EventSystem_AddEventCallback`

core/src/gateway_ll.c
ritave: @avranju Not right now, there's a team TODO task to separate Gateway init from starting up so that the callbacks can be registered.
This report is future proofing for now to ensure everything is logged correctly (also for potential side-effects, such as internal callbacks).

core/src/internal/event_system.c
ritave: @avranju No need, it's internal and the struct name itself is not anywhere inside the code

core/src/internal/event_system.c
ritave: @avranju No need, event_callbacks are always touched on the same, main, thread. Copy of event_callbacks is used for multi-threaded purpouses

core/src/internal/event_system.c
ritave: @avranju No need, see comment for lock @ AddEventCallbacks

core/src/internal/event_system.c
ritave: @avranju No need, the copy only happens inside the main Gateway thread.

core/src/internal/event_system.c
ritave: Probably yes, but the callback function requires a return value, even if the API itself requires call to exit so it's provided.

samples/callbacks_sample/src/main.c
ritave: It's not possible to listen to more events right now other than the destroy one. There's a future TODO to seperate Start and Init of Gateway so we can register for creation. Additionally, more events will be added in the future. Will change printed messages though.

PR #25, opened 2016-08-16
Add option to specify JDK_ARCH

srware:
Currently the JDK arch is hard-coded to 'amd64' which limits compatibility. This patch adds an option to specify JDK_ARCH using an environment variable and if not set it defaults to 'amd64'.

### DIFF
--- a/bindings/java/CMakeLists.txt
+++ b/bindings/java/CMakeLists.txt
@@ -12,11 +12,15 @@ if(
                       "JAVA_HOME to point to the location of the JDK.")
 endif()
 
-# Check if the variable JDK_ARCH exists and if not set it to 'amd64' by default
+# Check if the local or environment variable JDK_ARCH exists and if not set it to 'amd64' by default
 if(
     (NOT (DEFINED JDK_ARCH))
   )
-  set( JDK_ARCH amd64 )
+  if( DEFINED ENV{JDK_ARCH} )
+    set( JDK_ARCH $ENV{JDK_ARCH} )
+  else()
+    set( JDK_ARCH amd64 )
+  endif()
 endif()
 
 set(java_module_host_manager_sources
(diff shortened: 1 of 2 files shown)


andrew-buckley:
Hi @srware, thanks for having a look at this. I'll have a look and discuss it with the rest of the team shortly. 

From first glance it looks good! The only thing I could see potentially being an issue is setting environment variables on the users behalf in the CMakeLists.txt file unless it's really necessary. I'd suggest using a CMake variable to store the architecture, check if the environment variable exists and if it does set the variable to that, otherwise leave it at a default.

The other thing I'd say is to please update the Java docs in the repo to describe how users may change the default architecture.

Thanks again!

srware:
I agree, a CMake variable would be the preferred option. I will update the pull request and update the docs also.

Thanks.

avranju:
Also, I wonder if we cannot automatically determine what the CPU architecture is given that we have `JAVA_HOME`.

andrew-buckley:
I don't *think* so, the structure is identical (on linux) until you reach JAVA_HOME/lib/{architecture}. Additionally, the name of the JAVA_HOME could technically be anything

srware:
With that in mind what should take precedence, CMake variable or environment variable? I would say if someone specifies JDK_ARCH in their cmake command that should override environment?

avranju:
Agree. So, in summary, if `JDK_ARCH` is not defined and `ENV_{JDK_ARCH}` is, then assign `ENV_{JDK_ARCH}` to `JDK_ARCH`. Otherwise, assign `amd64` to `JDK_ARCH`.

srware:
The pull request commits have been updated to reflect the changes we talked about. Let me know what you think.

Thanks.

andrew-buckley:
Thanks, @srware. I'll be merging this in today.

PR #26, opened 2016-08-16
Ensure static libs cannot be built shared

srware:
Currently if the 'BUILD_SHARED_LIBS' flag is set to 'ON' static libraries are built as shared. As we are building both shared and static libraries for modules and bindings it makes sense to force library type so they can't be overridden.

### DIFF
--- a/bindings/dotnet/CMakeLists.txt
+++ b/bindings/dotnet/CMakeLists.txt
@@ -45,7 +45,7 @@ add_library(dotnet MODULE ${dotnet_sources}  ${dotnet_headers})
 target_link_libraries(dotnet gateway mscoree)
 
 #this builds the dotnet static library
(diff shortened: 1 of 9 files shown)


avranju:
Thanks for this PR @srware! This makes sense. I'd propose that we also explicitly build the main **gateway** library as a static library instead of relying on the CMake default. The change would be in [core/CMakeLists.txt](https://github.com/Azure/azure-iot-gateway-sdk/blob/develop/core/CMakeLists.txt#L81) on line 81.

srware:
Personally I like being able to specify how to build the libs depending on use case. Building the libs to be installed on a target device (as an rpm or deb package for example) is one scenario where shared libs should be used in my opinion.

avranju:
Agree. Have merged this PR via our CI gate.

PR #40, opened 2016-10-09
Device Ping tool

yaweiw:
Add device ping module and sample. The tool can verify the connectivity and reachability between the host and iot hub.


modules/iothubdeviceping/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+    ${iothubdeviceping_headers}
+)
+target_compile_definitions(iothubdeviceping_static PRIVATE BUILD_MODULE_TYPE_STATIC)
+target_link_libraries(iothubdeviceping_static
+    gateway
+    iothub_client
+)
+
+#this builds the IoTHubDevicePing_HL dynamic library (by default it uses IoTHub Module linked statically)
+add_library(iothubdeviceping_hl MODULE
```
damonbarry: We've recently made a breaking change that will affect you here...if you want your module to support a version of the `Module_Create` function that takes a JSON string instead of a struct, you no longer create a "high-level" (HL) module to do it. Instead, we added a new function to the module interface: `Module_CreateFromJson`. To migrate to the latest SDK, you'd do the following:
- remove the `iothubdeviceping_hl` module
- Implement the `Module_CreateFromJson` function in `iothubdeviceping` (use the code from `Module_Create` in your HL module)

See any of the modules in our latest `develop` branch for examples of this.

samples/device_ping/src/device_ping_lin.json
```diff
@@ -0,0 +1,29 @@
+{
+    "modules": [
+        {
+            "module name": "logger_hl",
```
damonbarry: Because we got rid of HL modules, references like this should be changed to something like:
```
    {
        "module name": "logger",
        "module path": "./modules/logger/liblogger.so",
        "args": {
            "filename": "log.txt"
        }
    },
```

damonbarry: ...same for the "win" version of the JSON config file.

modules/iothubdeviceping/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+    iothubdeviceping_static
+)
+
+if(${use_amqp})
+    target_link_libraries(iothubdeviceping_static iothub_client_amqp_transport)
+    target_link_libraries(iothubdeviceping iothub_client_amqp_transport)
+    linkUAMQP(iothubdeviceping_static)
+    linkUAMQP(iothubdeviceping)
+else()
+    add_definitions(-DIOTHUBMODULE_NULL_AMQP)
```
damonbarry: It looks like your sample is hard-coded to use HTTP and AMQP, so if `use_amqp` or `use_http` aren't set, you just need to print a message and fail cmake.

modules/iothubdeviceping/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+if(${use_http})
+    target_link_libraries(iothubdeviceping_static iothub_client_http_transport)
+    target_link_libraries(iothubdeviceping iothub_client_http_transport)
+    linkHttp(iothubdeviceping_static)
+    linkHttp(iothubdeviceping)
+else()
+    add_definitions(-DIOTHUBMODULE_NULL_HTTP)
+endif()
+
+if(${use_mqtt})
```
damonbarry: Your sample doesn't use MQTT so this block can be removed.

modules/iothubdeviceping/src/iothubdeviceping.c
```diff
@@ -0,0 +1,738 @@
...
+    {
+        result = malloc(sizeof(IOTHUBDEVICEPING_HANDLE_DATA));
+        if (result == NULL)
+        {
+            LogError("malloc returned NULL\r\n");
+        }
+        else
+        {
+            result->config = (IOTHUBDEVICEPING_CONFIG *)clone_iothub_device_ping_config(configuration);
+            result->iotHubClientHandle = IoTHubClient_LL_CreateFromConnectionString(((const IOTHUBDEVICEPING_CONFIG *)configuration)->DeviceConnectionString, HTTP_Protocol);
```
damonbarry: Why aren't you using the IoT Hub module? There's a lot of code in here that duplicates the logic in that module, and I don't think we should maintain two copies of it.

I think the device_ping sample gateway should load two modules: ping and iothub. The ping module publishes a ping message to the broker, and records the time when it was published. The iothub module receives the ping message and sends it to IoT Hub. The ping module also listens for events on the backing event hub and records the timestamp when the event arrives.

modules/iothubdeviceping/src/iothubdeviceping.c
```diff
@@ -0,0 +1,738 @@
...
+                message_receiver_context_arr[i].moduleHandleData = handleData;
+                message_receiver_context_arr[i].eventHubHost = eventHubHost;
+                message_receiver_context_arr[i].eventHubKeyName = eventHubKeyName;
+                message_receiver_context_arr[i].eventHubKey = eventHubKey;
+                message_receiver_context_arr[i].begin_execution_time = time(NULL);
+                message_receiver_context_arr[i].num_message_received = 0;
+                message_receiver_context_arr[i].message_received = false;
+                free(str);
+
+                if (ThreadAPI_Create(&message_receiver_context_arr[i].threadHandle, poll_eventhub_thread, &message_receiver_context_arr[i]) != THREADAPI_OK)
```
damonbarry: Messages sent on behalf of a particular device *always* land in the same event hub partition. So you only need to listen on one partition.

Check out the [ResolvePartitionIndex](https://github.com/Azure/azure-iot-sdks/blob/master/c/testtools/iothub_test/src/iothubtest.c#L225) function, used by the device SDK's end-to-end tests. If you pass to it the device ID (string) and the number of partitions in the hub, it returns the partition index.

modules/iothubdeviceping/src/iothubdeviceping.c
```diff
@@ -0,0 +1,738 @@
...
+    }
+    (void)Lock_Deinit(handleData->lockHandle);
+    free(handleData);
+}
+
+static void IoTHubDevicePing_Receive(MODULE_HANDLE moduleHandle, MESSAGE_HANDLE messageHandle)
+{
+    (void)moduleHandle;
+    (void)messageHandle;
+    // no action, IoTHubDevicePing is not interested in any messages
```
damonbarry: I'm beginning to wonder why this is a gateway sample, and not just a [device SDK](https://github.com/Azure/azure-iot-sdks) sample? It isn't doing anything gateway-specific.

samples/device_ping/src/main.c
```diff
@@ -0,0 +1,31 @@
...
+{
+    GATEWAY_HANDLE gateway;
+    if (argc != 2)
+    {
+        printf("usage: device_ping configFile\n");
+        printf("where configFile is the name of the file that contains the Gateway configuration\n");
+    }
+    else
+    {
+        if ((gateway = Gateway_Create_From_JSON(argv[1])) == NULL)
```
damonbarry: Breaking change: this function is now called `Gateway_CreateFromJson`.

doc/device_ping.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. A logger module
+2. A iot hub device ping module
+
+How it works:
+The iot hub device ping module will create a thread to send a message to the destination device previously provisioned in iot hub. Upon successful delivery, the message will be sinked to logger module for logging. The message delivery was done through iot hub API e.g. IoTHubClient_LL_SendEventAsync whose callback contains logic to poll the backend event hub to see if the message has arrived or delivery is timed out. The polling thread will simutaneously launch x number of worker threads to carry out the work for x number of event hub partitions. Whichever thread finds the wanted message will stop the rest from running.
+
+The logger module logs all the activities on the message broker to a file.
+
+## How to build the tool:
```
yaweiw: ok

doc/device_ping.md
```diff
@@ -0,0 +1,100 @@
...
+
+>run `ctest -C Debug -V`
+
+## How to run the tool:
+Linux
+
+- In a text editor, open the file `samples/device_ping/src/device_ping_lin.json` in your local copy
+of the azure-iot-gateway-sdk repository. You need to fill the following information:
+
+>`DeviceConnectionString`: Go to your iot hub 'Device Details' page and set the value to 'Connection string—primary key'.
```
yaweiw: roger that

modules/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ add_subdirectory(ble)
 add_subdirectory(simulated_device)
 add_subdirectory(identitymap)
 add_subdirectory(iothub)
+add_subdirectory(iothubdeviceping)
```
yaweiw: This makes great sense to me!

samples/device_ping/src/device_ping_lin.json
yaweiw: yes

samples/device_ping/src/main.c
yaweiw: yes fixed

modules/iothubdeviceping/src/iothubdeviceping.c
yaweiw: the purpose for this sample is to verify the gateway infrastructure is set up correct. yes it can be done as a device sdk sample but it won't serve that purpose.

modules/iothubdeviceping/src/iothubdeviceping.c
yaweiw: So here i tend to do it in a simpler way. Anyone can ping any device at any number of times and get the right time and results.

modules/iothubdeviceping/src/iothubdeviceping.c
yaweiw: the purpose for this sample is to verify the gateway infrastructure is set up correct. yes it can be done as a device sdk sample but it won't serve that purpose.

modules/iothubdeviceping/CMakeLists.txt
yaweiw: sure

modules/iothubdeviceping/CMakeLists.txt
yaweiw: sure

yaweiw:
@damonbarry you are right things the tool does can be definitely wrapped up in a device sdk sample. The original purpose of having this is to easily test the connectivity and reachability between a gw device and Iot hub. Even now in our happy path, after performing tons of setup work users will still be left in darkness without knowing if they've done it right until Lesson 4 - use Az Func to receive device-to-cloud messages. This doesn't sound right. We need a way to quickly and easily test the correctness of gw setup. That's what this comes from. I'd happy to see it as a tool/sample in our final release.

doc/device_ping.md
```diff
@@ -0,0 +1,100 @@
...
+
+>`EH_KEY_NAME`: Go to your iot hub 'Shared access policies' page and set the value to the policy name that has 'registry write', 'service connect' and 'device connect' permissions e.g. iothubowner
+
+>`EH_KEY`: The 'Primary key' of `EH_KEY_NAME`.
+
+>`EH_COMP_NAME`: Go to your iot hub 'Messaging' page and set the value to 'Event Hub-compatible name'.
+
+>`EH_PARTITION_NUM`: Go to your iot hub 'Messaging' page and set the value to 'Partitions'.
+
+Note: if you don't have D driver on your host machine, you'll need to change `filename` to other path.
```
damonbarry: I'd make device_ping_win.json consistent with device_ping_lin.json by setting "filename" to "log.txt". The current working directory is guaranteed to exist.  :)

yaweiw: makes sense

modules/iothubdeviceping/CMakeLists.txt
yaweiw: opps... sure

modules/iothubdeviceping/src/iothubdeviceping.c
yaweiw: Yes measuring time accurately for round trips is an integral part of what this tool should do. Since it sits in sample folder as hello_world does, we gotta demonstrate the use of GW infrastructure e.g. broker, module model don't we?:) A successful ping means two things: 1)GW sdk infra is all good 2)gw and iot hub are connected. And What is really interesting is when something breaks e.g. network drop, the output messages are helpful for diagnoses in most cases. We could potentially have a different implementation and put it under device sdk.

yaweiw:
@damonbarry

modules/iothubdeviceping/src/iothubdeviceping.c
damonbarry: "A successful ping means two things: 1)GW sdk infra is all good 2)gw and iot hub are connected"
Not true. You are creating a gateway with a custom module that talks to IoT Hub over HTTP. If the user intends to use our IoT Hub module (not custom), and the user wants to use MQTT or AMQP (not HTTP), then the sample doesn't give them confidence that their intended configuration will work.

yaweiw: Hi @damonbarry, thanks again for the great feedback. I've just added support for sending over AMQP and MQTT as suggested. we want the user to easily verify the correctness of their initial setup of the gateway environment as well as reachability of iot hub they've provisioned. Being able to send over the three protocols and confirm the arrival of the message with ease is useful to users. Please review. Thanks.

yaweiw:
@damonbarry

PR #50, opened 2016-11-15
display message flow for users

yaweiw:


damonbarry:
@yaweiw Can you paste a sample of the output before and after your changes here?  Thanks!

damonbarry:
Also, because this is a code contribution (as opposed to just docs) you'll want to base it on our `develop` branch instead of `master`.

PR #82, opened 2016-12-27
Update README.md

anhashia:
Add entrypoint for node_printer


aribeironovaes:
Hi Anil, 

Can you squash your commits? Looks like it would be a single commit. 

Thanks, 

Angelo Ribeiro.

Issue #96, opened 2017-01-12
Node.js module initialization issue when starting module is .NET


teamdevmatch:
Hey guys,

first of all I want to thank everybody who is contributing to this repo. I really think you're doing an amazing job here providing us with this open foundation for creating custom hybrid solutions in a fast and consistent way. Also looking forward to be a contributor very soon.

Right now I'm messing around with various gateway configurations and ran into an issue when setting up a Node.js module as sink and a starting .NET module as source.

When I start up the gateway it outputs the following message: **Info: Module has not been initialized yet: State: 1, Module: ../../../../samples/nodejs_simple_sample/nodejs_modules/filewriter.js**

Then the gateway prints: **Gateway is running.** and the .NET source module starts sending messages to the broker correctly.

The first message is not being processed by the Node module. 
Oddly enough, after ignoring the first message, the Node.js module starts processing incoming messages just fine.

I'm guessing the very first message kind of initializes the Node module and after that the module works as expected.

My question:
Is there anything that I'm missing in the configuration to instruct the gateway to initialize **all** modules on start? Am I doing something else wrong here or this an actual issue?

As a workaround I'm thinking about sending an initial wake-up / warm-up message, so that Node modules are initializing.

Thanks for reading all of that and y'all have a good one.

avranju:
Thanks for filing the issue @teamdevmatch. From what I can tell it sounds like a race condition of sorts. The .NET module appears to get initialized and starts producing messages before the Node.js module has had a chance to load and initialize. Are you commencing sending of messages from your .NET module from the `IGatewayModule.Create` method or are you using the `IGatewayModuleStart.Start` method as a signal for triggering the start of message production?

teamdevmatch:
Hey @avranju, the .NET module is the SDK's sensor sample (DotNetSensorModule) and the thread for publishing messages is created in the  `IGatewayModuleStart.Start` method. This should be the right place, I reckon.
Going to fiddle around a bit on replacing the `Thread` with a `Timer` or `Task` but I not sure if that'll make any difference.
Thanks so far looking into this topic.

avranju:
I see. I am going to say that this is an issue with the Node.js binding implementation. In this case I would expect the binding code that bridges the Node.js runtime with the rest of the gateway to delay delivery of the message to the JS code till it has been initialized. I'll file a task for this and get this addressed. Thanks for reporting this!

teamdevmatch:
[...]

PR #103, opened 2017-01-20
Added NuGet Package for .Net Framework target.

barnstee:
NuGet package contains everything you need to design & build a .Net-based GW module. Things that still need to be fixed in the SDK:
1. Build all .pdb files for DLLs in Release configuration
2. Build nanomsg DLL in Release configuration

### DIFF
--- a/nuget/Azure.IoT.Gateway.SDK.nuspec
+++ b/nuget/Azure.IoT.Gateway.SDK.nuspec
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+  <metadata>
+    <id>Azure.IoT.Gateway.SDK.Net</id>
+    <version>2017.01.13</version>
+    <title>Azure IoT Gateway SDK for .Net Framework</title>
+    <description>This package contains the Microsoft Azure IoT Gateway SDK targeting the .NET Framework</description>
+    <authors>Microsoft Corporation</authors>
+    <language>en-US</language>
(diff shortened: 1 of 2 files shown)


nuget/Azure.IoT.Gateway.SDK.nuspec
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+  <metadata>
+    <id>Azure.IoT.Gateway.SDK.Net</id>
+    <version>2017.01.13.2</version>
```
tameraw: >2017.01.13.2 [](start = 13, length = 12)

Investigate if this is supported by Nuget. I am not sure, and don't recall seeing this in packages I worked with.

barnstee: Yeah, I think it's needs to be 2017.01.13-2, checking...

barnstee: https://docs.microsoft.com/en-us/nuget/create-packages/prerelease-packages#semantic-versioning
So it needs to be 2017.01.13-alpha

zafields:
Address #96 and fixes the updates the build_nodejs.cmd to properly create the environment variables for the subsequent step on the node simple sample how to

damonbarry:
There's a much larger effort in process right now to produce gateway SDK packages for various platforms & languages, starting with NuGet/.NET. This will include our engineering process for publishing packages, so the actual .nuspec file is a relatively minor detail.

Since this PR has stalled a bit, we're going to close it. The work that was already scheduled (just getting underway now) will continue. Thanks!

PR #108, opened 2017-01-25
Sync node create

zafields:


zafields:
Addresses issue #96 by making the node binding create synchronous, and updates the `build_nodejs.cmd` to properly create the `NODE_INCLUDE` and `NODE_LIB` environment variables.

damonbarry:
@zafields there are build errors on [windows](http://azure-iot-sdks-ci.westus.cloudapp.azure.com:8080/job/new-win-c/2221//console).

bindings/nodejs/inc/nodejs_common.h
```diff
@@ -55,6 +56,7 @@ struct NODEJS_MODULE_HANDLE_DATA
     NODEJS_MODULE_HANDLE_DATA(NODEJS_MODULE_HANDLE_DATA&& rhs)
     {
         broker = rhs.broker;
+        create_complete = std::move(rhs.create_complete);
```
avranju: I am not sure, but did we not decide not to do this in the move constructor?

bindings/nodejs/src/nodejs.cpp
```diff
@@ -110,7 +112,31 @@ static MODULE_HANDLE NODEJS_Create(BROKER_HANDLE broker, const void* configurati
                 }
                 else
                 {
-                    handle_data->SetModuleState(NodeModuleState::initializing);
+                    // Wait for the module to become fully initialized
+                    NodeModuleState module_state;
+                    std::future<NodeModuleState> create_gate = handle_data->create_complete.get_future();
```
avranju: Use `auto` instead of spelling out the type?

bindings/nodejs/src/nodejs.cpp
```diff
@@ -110,7 +112,31 @@ static MODULE_HANDLE NODEJS_Create(BROKER_HANDLE broker, const void* configurati
...
+                    switch (create_gate.wait_for(std::chrono::seconds(10)))
```
avranju: Define a constant for the timeout period of `10` seconds? In fact the constant could perhaps be a `std::chrono` instance.

bindings/nodejs/src/nodejs.cpp
```diff
@@ -110,7 +112,31 @@ static MODULE_HANDLE NODEJS_Create(BROKER_HANDLE broker, const void* configurati
...
+                    {
+                      case std::future_status::ready:
+                          module_state = create_gate.get();
+                        break;
```
avranju: Nit: `break` indentation should align with line 121. Here and on line 127.

bindings/nodejs/inc/nodejs_common.h
```diff
@@ -140,16 +142,17 @@ struct NODEJS_MODULE_HANDLE_DATA
...
+    BROKER_HANDLE broker;
+    std::promise<NodeModuleState> create_complete;
+    std::string main_path;
+    std::string configuration_json;
+    v8::Isolate *v8_isolate;
+    v8::Persistent<v8::Object> module_object;
+    size_t module_id;
+    PFNMODULE_START on_module_start;
+    NodeModuleState module_state;
+    bool start_pending;
```
avranju: The idea behind having this flag `start_pending` was to handle the case where a call to `NODEJS_Start` occurs before the module's state has transitioned to `initialized` from `initializing`. Now that `NODEJS_Create` is a blocking call and we no longer have an `initializing` state, we may not need this flag anymore. Consider removing this field and the associated logic.

tools/build_nodejs.cmd
```diff
@@ -47,8 +47,7 @@ copy node\build\Release\lib\*.lib dist\lib
 
 popd
 
-rem Export environment variables for where the include/lib files can be found
-set "NODE_INCLUDE=%build-root%\dist\inc"
-set "NODE_LIB=%build-root%\dist\lib"
+rem Set environment variables for where the include/lib files can be found
+@endlocal & set NODE_INCLUDE=%build-root%\dist\inc\ & set NODE_LIB=%build-root%\dist\lib\
```
avranju: Nice!

bindings/nodejs/src/nodejs.cpp
zafields: fixed (overly helpful VS caused this).

bindings/nodejs/src/nodejs.cpp
zafields: done.

bindings/nodejs/src/nodejs.cpp
zafields: Just gave the number a name for now.

tools/build_nodejs.cmd
zafields: 👍

bindings/nodejs/inc/nodejs_common.h
zafields: Regardless of the life of any individual data structure used to contain the handle data, the collection of handle data represents a single object. So I believe the future (along with all other data) should be moved if possible to reflect the intent of what is happening.
Unfortunately, we are only able to preserve this data in the MOVE constructor and not in the copy constructors. I see how this is more confusing than not bring them along all together, but I must admit I don't like the way it feels.

bindings/nodejs/inc/nodejs_common.h
zafields: removed with test passing

bindings/nodejs/inc/nodejs_common.h
zafields: I'm going to leave a comment to this effect and forget about it. :D

PR #110, opened 2017-01-27
Add test for NODEJS_Create

zafields:
Test confirms that NODEJS_Create is synchronous, which is implicitly required by the current design of the IoT Field Gateway

### DIFF
--- a/bindings/nodejs/tests/nodejs_int/nodejs_int.cpp
+++ b/bindings/nodejs/tests/nodejs_int/nodejs_int.cpp
@@ -303,16 +303,16 @@ static void publish_mock_message(const v8::FunctionCallbackInfo<v8::Value>& info
 }
 
 static const char * NOOP_JS_MODULE = "module.exports = { "   \
-"create: function(broker, configuration) { "       \
-"  console.log('create'); "                            \
-"  return true; "                                      \
-"}, "                                                  \
-"receive: function(msg) { "                            \
-"  console.log('receive'); "                           \
-"}, "                                                  \
-"destroy: function() { "                               \
-"  console.log('destroy'); "                           \
-"} "                                                   \
+"create: function(broker, configuration) { "                 \
+"  console.log('create'); "                                  \
+"  return true; "                                            \
+"}, "                                                        \
+"receive: function(msg) { "                                  \
+"  console.log('receive'); "                                 \
+"}, "                                                        \
+"destroy: function() { "                                     \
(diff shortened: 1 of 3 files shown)


bindings/nodejs/tests/nodejs_int/nodejs_int.cpp
```diff
@@ -1292,4 +1292,77 @@ BEGIN_TEST_SUITE(nodejs_int)
...
+
+        MESSAGE_HANDLE message = Message_Create(&msg_config);
+        ASSERT_IS_NOT_NULL(message);
+
+        auto result = NODEJS_Create(g_broker, &config);
+        NODEJS_MODULE_HANDLE_DATA* handle_data = reinterpret_cast<NODEJS_MODULE_HANDLE_DATA*>(result);
+        ASSERT_IS_NOT_NULL(handle_data);
+
+        ///act
+        NODEJS_Receive(handle_data, message);
```
darobs: That's so straightforward a test, I like it.

bindings/nodejs/tests/nodejs_int/nodejs_int.cpp
```diff
@@ -303,16 +303,16 @@ static void publish_mock_message(const v8::FunctionCallbackInfo<v8::Value>& info
...
+"create: function(broker, configuration) { "                 \
+"  console.log('create'); "                                  \
+"  return true; "                                            \
+"}, "                                                        \
+"receive: function(msg) { "                                  \
+"  console.log('receive'); "                                 \
+"}, "                                                        \
+"destroy: function() { "                                     \
+"  console.log('destroy'); "                                 \
+"} "                                                         \
```
avranju: Is the formatting of this string constant off?

zafields: This is purely cosmetic. I noticed that slashes seemed to be in a column, so I just put them all in alignment.

zafields:
@damonbarry This check in is gated on a bug fix that Raj caught, fixed and has out for code review in CodeFlow. I have checked this test against the negative case to prove it is actually testing something and is otherwise ready to go.

avranju:
@zafields The bug fix has been [checked in](https://github.com/Azure/azure-iot-gateway-sdk/commit/a8a9d8e66e65c0226bfd7a9265e01fd1b8c2c703). You should be able to push your changes now.

Issue #120, opened 2017-02-06
build.sh --clean doesn't work


GrantEdwards:
The build.sh --clean option appears to be ignored.  There doesn't seem to be any way to rebuild
sources.

damonbarry:
Yeah, `--clean` was part of the original build script, but didn't see much use and eventually went stale. The right thing to do is to remove it, since the build script deletes the cmake `build/` folder every time (i.e., every build is a clean build).

If you've already built and you want to rebuild the gateway bits without calling the build script, the way to do it in Linux is (from the repo root folder):
```
cd build/
make clean
make
```

If you're talking about rebuilding everything including dependencies, there's a flag in the latest build script for that: `--rebuild-deps`.

damonbarry:
Fix is in the [2017-03-06](https://github.com/Azure/azure-iot-gateway-sdk/tree/2017-03-06) release.

PR #123, opened 2017-02-07
Remove unused build script option

damonbarry:
The `--clean' option in `tools/build.sh` wasn't doing anything, as @GrantEdwards pointed out in issue #120.

### DIFF
--- a/tools/build.sh
+++ b/tools/build.sh
@@ -4,7 +4,6 @@
 
 set -e
 
-build_clean=
 build_root=$(cd "$(dirname "$0")/.." && pwd)
 local_install=$build_root/install-deps
 log_dir=$build_root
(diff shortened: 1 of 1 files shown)


tools/build.sh
```diff
@@ -4,7 +4,6 @@
 
 set -e
 
-build_clean=
 build_root=$(cd "$(dirname "$0")/.." && pwd)
```
dcristoloveanu: >d_r [](start = 4, length = 3)

It would actually be so nice if this script would die a horrible death.
I think all the options handled by this script can be directly passed to cmake, isn't it. It seems this script is an extra abstraction layer that is not really necessary ...

dcristoloveanu:
:shipit:

tools/build.sh
damonbarry: Totally agree, and I think that will happen in the not too distant future.

PR #124, opened 2017-02-07
Only remove links if they weren't already removed

damonbarry:
In the course of investigating #86 I updated the submodule reference to azure-c-shared-utility locally and noticed new error spew in the console when running our gateway samples:
```
Error: Time:Tue Jan 31 18:10:21 2017 File:D:\tmp\azure-iot-gateway-sdk\deps\c-utility\src\vector.c Func:_VECTOR_size Line:328 invalid argument handle(NULL).
Error: Time:Tue Jan 31 18:10:21 2017 File:D:\tmp\azure-iot-gateway-sdk\deps\c-utility\src\vector.c Func:_VECTOR_find_if Line:291 invalid argument - handle(00000000), pred(50FB8280)
Error: Time:Tue Jan 31 18:10:21 2017 File:D:\tmp\azure-iot-gateway-sdk\deps\c-utility\src\vector.c Func:_VECTOR_size Line:328 invalid argument handle(NULL).
Error: Time:Tue Jan 31 18:10:21 2017 File:D:\tmp\azure-iot-gateway-sdk\deps\c-utility\src\vector.c Func:_VECTOR_find_if Line:291 invalid argument - handle(00000000), pred(50FB8280)
```
I discovered that these error messages were recently added to the VECTOR library. They are innocuous in this case, but they point to the fact that we're trying to remove links twice:

In `gateway_destroy_internal`, we loop through all the links and call `gateway_removelink_internal` for each one. Then we loop through all the modules and call `gateway_removemodule_internal` for each one. Inside `gateway_removemodule_internal`, we try to loop through all the links again and call `gateway_removelink_internal` *again* if the module we’re removing is listed either as the source or the sink for that link.

To fix this, I test for a NULL `links` VECTOR before entering the while loop that would try (in vain) to call `gateway_removelink_internal` a second time. I can't remove the while loop because the parent function (`gateway_removemodule_internal`) is called from a few other code paths as well; not all paths call `gateway_removelink_internal` twice. So the NULL test is a reasonable solution.

### DIFF
--- a/core/src/gateway_internal.c
+++ b/core/src/gateway_internal.c
@@ -554,8 +554,13 @@ void gateway_removemodule_internal(GATEWAY_HANDLE_DATA* gateway_handle, MODULE_D
     remove_module_from_any_source(gateway_handle, *module_data_pptr);
     LINK_DATA *link;
     /* Codes_SRS_GATEWAY_26_018: [ This function shall remove any links that contain the removed module either as a source or sink. ] */
-    while ((link = VECTOR_find_if(gateway_handle->links, link_name_both_find, (*module_data_pptr)->module_name)) != NULL)
-        gateway_removelink_internal(gateway_handle, link);
+    if (gateway_handle->links)
+    {
+        while ((link = VECTOR_find_if(gateway_handle->links, link_name_both_find, (*module_data_pptr)->module_name)) != NULL)
+        {
+            gateway_removelink_internal(gateway_handle, link);
+        }
+    }
 
     free((*module_data_pptr)->module_name);
 
--- a/core/tests/gateway_createfromjson_ut/gateway_createfromjson_ut.cpp
+++ b/core/tests/gateway_createfromjson_ut/gateway_createfromjson_ut.cpp
@@ -980,14 +980,6 @@ TEST_FUNCTION(Gateway_Create_Start_fails_returns_null)
     STRICT_EXPECTED_CALL(mocks, VECTOR_erase(IGNORED_PTR_ARG,IGNORED_PTR_ARG, 1))
         .IgnoreArgument(1)
         .IgnoreArgument(2);
-    STRICT_EXPECTED_CALL(mocks, VECTOR_find_if(IGNORED_PTR_ARG, IGNORED_PTR_ARG, IGNORED_PTR_ARG))
-        .IgnoreArgument(1)
-        .IgnoreArgument(2)
-        .IgnoreArgument(3);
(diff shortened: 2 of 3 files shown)


core/src/gateway_internal.c
```diff
@@ -554,8 +554,11 @@ void gateway_removemodule_internal(GATEWAY_HANDLE_DATA* gateway_handle, MODULE_D
     remove_module_from_any_source(gateway_handle, *module_data_pptr);
     LINK_DATA *link;
     /* Codes_SRS_GATEWAY_26_018: [ This function shall remove any links that contain the removed module either as a source or sink. ] */
-    while ((link = VECTOR_find_if(gateway_handle->links, link_name_both_find, (*module_data_pptr)->module_name)) != NULL)
-        gateway_removelink_internal(gateway_handle, link);
+    if (gateway_handle->links)
```
dcristoloveanu: >(gateway_handle->links) [](start = 7, length = 23)

/cough unit test?

dcristoloveanu:
:shipit:

core/src/gateway_internal.c
```diff
@@ -554,8 +554,11 @@ void gateway_removemodule_internal(GATEWAY_HANDLE_DATA* gateway_handle, MODULE_D
...
+    {
+        while ((link = VECTOR_find_if(gateway_handle->links, link_name_both_find, (*module_data_pptr)->module_name)) != NULL)
+            gateway_removelink_internal(gateway_handle, link);
```
avranju: Nit: Braces around control blocks even when they have only 1 line?

damonbarry: I can add that, no problem. I'm also having to update some unit tests...

core/src/gateway_internal.c
damonbarry: I'll think about what unit test(s) we can/should add for this...

damonbarry: This is an internal function that is part of the code path for five public functions. Some enter the `if` clause and some don't. The result is effectively the same before and after this change, except now we don't print out misleading LogError statements. I don't see much value in a targeted unit test for this, beyond what we already have.

nielsms:
@damonbarry I think you should also test for NULL gateway_handle->links around the:
    
remove_module_from_any_source(gateway_handle, *module_data_pptr);

line in the gateway_removemodule_internal function (line 554).
This method also calls VECTOR_size on the links vector after it has been destroyed, leading to the error when the gateway exits.

damonbarry:
@nielsms I've been meaning to track that one down...thanks for the info! I'll get a fix in soon.

PR #128, opened 2017-02-10
Out-of-process Module design and API

zafields:


core/devdoc/oop_module_requirements.md
```diff
@@ -0,0 +1,106 @@
...
+void
+OopModule_Detach (
+    OOP_MODULE_HANDLE oop_module_
+);
+```
+
+### OopModule_DoWork
+
+`OopModule_DoWork` is intended to provide the caller with fine-grain control of the
+worker thread. Therefore, `OopModule_DoWork` is a non-blocking call, which will check
```
damonbarry: DoWork doesn't give the caller fine-grained control over the worker thread--there may not even *be* a worker thread. I'd say that `OopModule_DoWork` is intended to be called in situations where users have opted not to use the start/stop convenience functions. Make it clear that these are mutually exclusive. Users should call DoWork themselves, or use the start/stop functions, but not both.

core/devdoc/oop_module_requirements.md
```diff
@@ -0,0 +1,106 @@
...
+listening for the gateway create message. Once the create message is received
+the OopModule library will call `Module_ParseConfigurationFromJson`^, `Module_Create`
+and `Module_FreeConfiguration`^.
+*^ Assuming the optional API has been provided.*
+
+```c
+extern
+OOP_MODULE_HANDLE
+OopModule_Attach (
+    const MODULE_API * module_api_,
```
damonbarry: I've used variable names with trailing underscores in C++ for private/protected class members. So there's nothing unusual about seeing it here, *except* that you won't find that pattern anywhere in our SDKs. I recommend you follow the existing patterns in this regard.

That said, we're rather inconsistent with variable names in public functions. I've seen `boxcar_casing` and `camelCasing`. You pick.

core/devdoc/oop_module_requirements.md
```diff
@@ -0,0 +1,106 @@
...
+extern
+void
+OopModule_HaltWorkerThread (
+    OOP_MODULE_HANDLE oop_module_
+);
+```
+
+### OopModule_StartWorkerThread
+
+`OopModule_StartWorkerThread` is a convenience method which gives control of
```
damonbarry: I assume these paragraphs will be turned coded requirements, complete with IDs and the formal language (e.g., "shall") we all adore?

core/src/oop_module.c
```diff
@@ -0,0 +1,308 @@
...
+    const char * const connection_key_
+) {
+    // Allocate and initialize the out-of-process module data structure
+    OOP_MODULE_HANDLE oop_module = (OOP_MODULE_HANDLE)calloc(1, sizeof(OOP_MODULE));
+
+    // Save the module API
+    oop_module->module->module_apis = module_api_;
+
+    // Connect to the message server
+    connect_module_to_message_server_with_key(oop_module, connection_key_);
```
damonbarry: This function call should be `connect_to_gateway_with_connection_key`, right?

core/src/oop_module.c
```diff
@@ -0,0 +1,308 @@
...
+
+    return oop_module;
+}
+
+void
+OopModule_Detach (
+    OOP_MODULE_HANDLE oop_module_
+) {
```
damonbarry: TODO

core/src/oop_module.c
```diff
@@ -0,0 +1,308 @@
...
+    // Send termination signal
+    nn_term(); //TODO: Learn more about this vs. nn_close
+
+    free(oop_module_);
+    return;
+}
+
+// NOTE: Message pump should not block (allows message pump to be called from a single-threaded application).
+void
+OopModule_ManualMessagePump (
```
damonbarry: `OopModule_DoWork`

core/src/oop_module.c
```diff
@@ -0,0 +1,308 @@
...
+    int message_socket;
+    MESSAGE_THREAD_HANDLE message_thread;
+    MODULE module;
+} OOP_MODULE;
+
+OOP_MODULE_HANDLE
+OopModule_Attach (
+    const MODULE_API * module_api_,
+    const char * const connection_key_
+) {
```
damonbarry: I assume error handling for public functions will come in another commit added to this PR...

core/devdoc/oop_module_requirements.md
```diff
@@ -0,0 +1,106 @@
...
+```c
+typedef OOP_MODULE * OOP_MODULE_HANDLE;
+```
+
+## Exposed API
+
+### OopModule_Attach
+
+`OopModule_Attach` will provide the OopModule library with the caller's
+"Module API" and the connection key required to communicate with the Azure IoT
```
zafields: You sir, are a poet of code. I have moved to token - fixed in coming commit.

core/devdoc/oop_module_requirements.md
```diff
@@ -0,0 +1,106 @@
...
+Field Gateway. When `OopModule_Attach` is called the OopModule library will begin
```
zafields: I'm not sure where I heard this name from. Probably those seedy guys I hang out with down by the loading dock... Or it's my failed and thinly veiled attempt to claim all the glory by branding our product as the Fields' Gateway...

Removed in the upcoming commit.

core/devdoc/oop_module_requirements.md
zafields: Removed in upcoming commit.

core/devdoc/oop_module_requirements.md
zafields: I was just going to jump straight into *Olde English*, but I thought I would lose the audience.

core/inc/oop_module.h
```diff
@@ -0,0 +1,38 @@
+
+#ifndef OOP_MODULE_H
+#define OOP_MODULE_H
+
+typedef OOP_MODULE * OOP_MODULE_HANDLE;
+
+extern
```
darobs: Are we exporting this from the library this will reside in?  Do we need a macro like `GATEWAY_EXPORT` to declare these functions?

(Also, have we given thought about how this library will be delivered?  static or dynamic, etc...)

core/devdoc/oop_module_requirements.md
damonbarry: We used to call it that, pre-GA. Then "Field" got dropped. So it wasn't all in your head...

core/devdoc/oop_module_requirements.md
zafields: I have amended the description in the upcoming commit. I think I did better, but I'll let you be the judge.

core/src/oop_module.c
zafields: That will come with the TDD based on the finalized requirements

core/src/oop_module.c
zafields: I was playing around with the notion of splitting out this functionality. but it turns out to be straight-forward enough that I don't see any value to having it as a static function.

core/src/oop_module.c
zafields: fixed in upcoming commit

core/inc/oop_module.h
zafields: I have added the appropriate decorations (i.e. MOCKABLE_FUNCTION and GATEWAY_EXPORT).

core/devdoc/oop_module_requirements.md
ancaantochi: I think a descriptive name should be used instead of DoWork and StartWorkerThread,  something like ReceiveOOPMessage and StartListening.

damonbarry: @ancaantochi Your names are good suggestions because they avoid implementation details like data types or sync primitives. However, DoWork is consistent with the C device SDK, and the worker thread sends as well as receives, so "listening" isn't entirely accurate.

zafields: @damonbarry As it stands, `DoWork` only receives messages. Broker_Publish is intercepted and directly invokes nanomsg, so it falls outside of `DoWork`. Do you have a vision for how this would work? Feels like a queueing mechanism.

zafields:
@damonbarry @darobs @ancaantochi I have updated this PR to be design and API only

zafields:
Merged into @darobs/oop-modules

PR #134, opened 2017-02-17
[REVIEW REQUESTED] Oop module preliminary implementation

zafields:


core/src/oop_module.c
```diff
@@ -0,0 +1,720 @@
...
+    const char * json_config
+);
+
+int
+worker_thread (
+    void * thread_arg
+);
+
+int
+process_module_create_message (
```
damonbarry: I see the word "process" used in this file two different ways:
1. Out-of-**process** module (an instance of a program)
2. `process_module_create_message` and `invoke_add_module_process` (set of activities that achieve a result)

For (2) can we call it something else to avoid confusion? Something like `handle_module_create_message` and `init_module`...

core/src/oop_module.c
```diff
@@ -0,0 +1,720 @@
...
+                LogError("%s: Unable to parse control message!", __FUNCTION__);
+            } else {
+                // Route control channel messages to appropriate functions
+                switch (structured_control_message->type) {
+                  case CONTROL_MESSAGE_TYPE_MODULE_CREATE:
+                    if (0 != process_module_create_message(oop_module, (CONTROL_MESSAGE_MODULE_CREATE *)structured_control_message)) {
+                        LogError("%s: Unable to process create message!", __FUNCTION__);
+                    }
+                    break;
+                  case CONTROL_MESSAGE_TYPE_MODULE_START: if (((MODULE_API_1 *)oop_module->module.module_apis)->Module_Start) { ((MODULE_API_1 *)oop_module->module.module_apis)->Module_Start(oop_module->module.module_handle); } break;
```
damonbarry: ...just a _little_ too much code for one line. 🤷‍♂️

core/src/oop_module.c
```diff
@@ -0,0 +1,720 @@
...
+            (void)nn_freemsg(control_message);
+        }
+
+        // Check first message on each message channel
+        for (int i = 0; i < oop_module->message_channel_count ; ++i) {
+            void * module_message = NULL;
+
+            bytes_received = nn_recv(oop_module->message_sockets[i], &module_message, NN_MSG, NN_DONTWAIT);
+            if (bytes_received < 0 && EAGAIN != bytes_received) {
+                // unexpected error
```
damonbarry: Are you going to handle this somehow?

core/inc/oop_module.h
```diff
@@ -0,0 +1,30 @@
...
+
+#ifdef __cplusplus
+  extern "C" {
+#endif
+
+typedef struct OOP_MODULE_TAG * OOP_MODULE_HANDLE;
+
+#include "azure_c_shared_utility/umock_c_prod.h"
+
+MOCKABLE_FUNCTION(, GATEWAY_EXPORT OOP_MODULE_HANDLE, OopModule_Attach, const MODULE_API *, module_apis, const char * const, connection_id);
```
zafields: removed extraneous `const`

core/src/oop_module.c
```diff
@@ -0,0 +1,720 @@
...
+#include "control_message.h"
+#include "gateway.h"
+#include "message.h"
+
+typedef enum OOP_MODULE_STATUS_TAG {
+    OOP_MODULE_DETACH = -1,
+    OOP_MODULE_OK,
+    OOP_MODULE_GATEWAY_CONNECTION_ERROR,
+    OOP_MODULE_MODULE_CREATION_ERROR,
+} OOP_MODULE_STATUS;
```
zafields: changed

core/src/oop_module.c
zafields: I feel like "handle" is an ironic choice, but I agree that we need a better name.

core/src/oop_module.c
zafields: but it's clearly marked that I don't expect it to happen... 😜

zafields: fixed.

core/src/oop_module.c
zafields: I'm sure they told Picaso that both eyes shouldn't be on the same sides of the head...

fixed.

core/src/oop_module.c
```diff
@@ -0,0 +1,720 @@
...
+        LogError("%s: NULL parameter - connection_id", __FUNCTION__);
+        oop_module = NULL;
+    } else if (GATEWAY_CONNECTION_ID_MAX > strnlen(connection_id, (GATEWAY_CONNECTION_ID_MAX + 1)) ) {
+        LogError("%s: connection_id exceeds GATEWAY_CONNECTION_ID_MAX in length", __FUNCTION__);
+        oop_module = NULL;
+    // Allocate and initialize the out-of-process module data structure
+    } else if (NULL == (oop_module = (OOP_MODULE_HANDLE)calloc(1, sizeof(OOP_MODULE)))) {
+        LogError("%s: Unable to allocate memory!", __FUNCTION__);
+    } else {
+        static const size_t ENDPOINT_LENGTH_MAX = (sizeof("ipc://") + NN_SOCKADDR_MAX);
```
darobs: >NN_SOCKADDR_MAX [](start = 70, length = 15)

I put ".ipc" at the end of the URI in my code, so we should probably come to an agreement on that.

zafields: I don't know much about `nanomsg`, so I may not be doing it right? Is `.ipc` mandatory, or is it a this-or-that between `.ipc` and `ipc://`, or is it superfluous? If it's mandatory, then I will switch immediately. If it's a this or that, I'm good with your choice. However, if it's a decorator, then I think we should leave it off as it may cause confusion if we were to use an alternative prefix.

darobs: I don't remember which it was.  I added the ".ipc" extension during the spike, but don't recall why.  I don't think nanomsg requires it.

darobs: I will remove it in my branch, and we'll see what happens on integration.

zafields: Side discussion with @darobs resulted in temporarily dropping the extension, until we can prove that we do or do not need it.

damonbarry:
Moving this work into a feature branch.

PR #135, opened 2017-02-23
Avoid needless VECTOR_size errors

damonbarry:
This is a follow-up to #124. There were a few more VECTOR_size errors that we can avoid by testing for `gateway_handle->links == NULL`.

@nielsms

### DIFF
--- a/core/src/gateway_internal.c
+++ b/core/src/gateway_internal.c
@@ -552,10 +552,10 @@ void gateway_removemodule_internal(GATEWAY_HANDLE_DATA* gateway_handle, MODULE_D
     module.module_handle = (*module_data_pptr)->module;
 
     remove_module_from_any_source(gateway_handle, *module_data_pptr);
-    LINK_DATA *link;
     /* Codes_SRS_GATEWAY_26_018: [ This function shall remove any links that contain the removed module either as a source or sink. ] */
     if (gateway_handle->links)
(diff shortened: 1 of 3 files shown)


core/src/gateway_internal.c
```diff
@@ -690,23 +690,26 @@ int add_module_to_any_source(GATEWAY_HANDLE_DATA* gateway_handle, MODULE_DATA* m
...
-            if (module_sink == NULL)
+            LINK_DATA * link_data = VECTOR_element(gateway_handle->links, link);
+            if (link_data->from_any_source)
             {
-                LogError("Could not find sink for link [%s]", link_data->module_sink);
-            }
-            else
-            {
-                if (remove_one_link_from_broker(gateway_handle, module->module, (*module_sink)->module) != 0)
+                MODULE_DATA** module_sink = (MODULE_DATA**)VECTOR_find_if(gateway_handle->modules, module_name_find, link_data->module_sink->module_name);
```
aribeironovaes: Do we need to test if link data is null?

damonbarry: I don't think so. If `gateway_handle->links` is non-NULL, and VECTOR_size returns a non-zero value, then the vector element will be valid. We maintain this vector internally, so we can guarantee consistency/integrity.

PR #136, opened 2017-02-23
Updated dotnet-binding project…

jeffreyCline:
Updated dotnet-binding project files to support Release_Delay_Sign build configuration for SN signing of the .NET binding dlls.


jeffwilcox:
Microsoft employee confirmed, process as needed; we'll need to investigate the underlying CLA issue.

avranju:
Just curious, why do we want these assemblies strong named? The [general guidance](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/strong-name-signing.md#1-microsoft-strong-names-their-assemblies-should-i) that I have seen is to err on the side of not strong naming assemblies.

jeffreyCline:
We were attempting to mirror the existing Azure Client SDK code sign practices. I agree, having the assemblies SN can cause issues and I would actually prefer not to do it, but as I said we were trying to implement exactly what the client team is already doing.

PR #146, opened 2017-03-03
Clarify Microsoft.Azure.IoT.Gateway's public interface

damonbarry:
- Mark implementation details as `internal`
- In build_dotnet_core.cmd, exit with an error if tests fail

I confirmed that unit tests pass and the sample still works.


aribeironovaes:
:shipit:

PR #147, opened 2017-03-04
Some refactoring before the 1st release of the .NET Core binding

damonbarry:
This touches a moderate amount of code, but each commit is well-scoped. Here are some details:

### Why change the namespace? And why not change it for the .NET Framework binding too?
- Since the .NET Core binding is new, we might as well conform to the namespace used by the device SDK (Microsoft.Azure.Devices...)
- This change will also prevent type clashes in gateways that run .NET Framework *and* .NET Core modules.
- We can bring the .NET Framework binding over to this namespace in the future, and reuse much of the code between the two bindings.

### Why make the interop argument to the Broker constructor explicit? And why make it an interface instead of a concrete class?
- `new Broker(...)` is only called internally, so the extra argument isn't a problem
- We previously hid the interop argument because it was only for unit testing, but in the future we'll be able to share the broker code between .NET Framework and .NET Core bindings, and just pass in a different interop object

### What is global.json for?
- The .NET Core sample modules live with the sample now, so they're no longer siblings to Microsoft.Azure.Devices.Gateway, which they depend on. A global.json file at the common parent folder (in this case, the repo root) ensures `dotnet restore` can find the dependency.

### What did you do to the build_dotnet_core build scripts?
- The bash script was using the `dotnet` commands to build, while the windows script was using msbuild. I changed the windows script to use `dotnet` commands like the bash script, and improved the arguments that can be passed to both scripts.


damonbarry:
@Azure/iot-gateway-devs FYI

aribeironovaes:
:shipit:

damonbarry:
FYI I accidentally made some changes in CMakeLists.txt to the .NET Framework stuff (was only supposed to touch the .NET Core stuff). Jenkins caught the error. I fixed the commit `Move .NET Core binding to Microsoft.Azure.Devices namespace` and cherry-picked the subsequent commits. I'll try Jenkins again...

PR #148, opened 2017-03-06
Nuget config

shsushi:
Some Nuget packages(Castle.Core.3.3.3, Moq.4.5.10) are only in old nuget feed, needs to provide a complete feed URLs to avoid download failures.


damonbarry:
These changes we merged into `master`.

PR #154, opened 2017-03-09
Any CPU building issue for dotnet

shsushi:


aribeironovaes:
2 Overall comments: 
-We should rebase to master, instead of merging, so we avoid extra commits on this process; 
-We need to make sure we test this End2End as well as in our build script. For that we should wait for the update It's being done to the latest .NET. So, I suggest we hold this pull request once this change is in.

aribeironovaes:
Ok, forget about comment 2. I just realized that this is for .NET, not ,NET Core. 

So, the only think I can say is that this needs to be rebased into master and have a single commit. So, please, Rebase and Squash commits so we can merge into master...

jeffreyCline:
There are other changes that are needed as well, the code sign script, build nuget script, and nuspec files need to be updated to support Any CPU as well.

shsushi:
@aribeironovaes , yes i will do a rebase.

PR #157, opened 2017-03-10
Update dev Dockerfile

damonbarry:
I created this from scratch a week or two ago. It turns out our current Dockerfile is out-of-date, so I'm proposing we replace it with this.

It's based on Ubuntu 16.04 instead of 14.04. And it builds the Node.js, Java, and .NET Core bindings.

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -1,42 +1,46 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
 
-RUN apt-get clean && \
-    apt-get update && \
-    apt-get install -y \
-        software-properties-common && \
-    apt-add-repository ppa:webupd8team/java && \
-    echo oracle-java8-installer shared/accepted-oracle-license-v1-1 select true | sudo /usr/bin/debconf-set-selections && \
-    apt-get update && \
-    apt-get install -y \
-        oracle-java8-installer \
-        oracle-java8-set-default \
-        maven \
-        python2.7 \
-        curl \
-        build-essential \
-        libcurl4-openssl-dev \
-        git \
-        cmake \
-        libssl-dev \
-        uuid-dev \
-        valgrind \
-        libglib2.0-dev
+RUN apt-get update && apt-get install -y apt-transport-https
 
-COPY build_nodejs.sh /tools/
+RUN sh -c 'echo "deb [arch=amd64] https://apt-mo.trafficmanager.net/repos/dotnet-release/ xenial main" > /etc/apt/sources.list.d/dotnetdev.list' \
+    && apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 417A0893 \
+    && apt-get update
 
-RUN /tools/build_nodejs.sh
+RUN apt-get install -y \
+    curl \
+    build-essential \
+    openjdk-8-jdk \
+    git \
+    cmake \
+    maven \
+    valgrind \
+    pkg-config \
+    uuid-dev \
+    libcurl4-openssl-dev \
+    libssl-dev \
+    libglib2.0-dev \
+    dotnet-dev-1.0.0-preview2.1-003177
 
-ENV JAVA_HOME /usr/lib/jvm/java-8-oracle
+WORKDIR /
 
-ENV NODE_INCLUDE /build_nodejs/dist/inc
-ENV NODE_LIB /build_nodejs/dist/lib
+RUN git clone https://github.com/Azure/azure-iot-gateway-sdk.git
 
-ENV GATEWAY_REPO https://github.com/Azure/azure-iot-gateway-sdk
-ENV COMMIT_ID master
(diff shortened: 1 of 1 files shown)


damonbarry:
@Azure/iot-gateway-devs

tools/Dockerfile
```diff
@@ -1,42 +1,46 @@
...
+    dotnet-dev-1.0.0-preview2.1-003177
+
+WORKDIR /
+
+RUN git clone https://github.com/Azure/azure-iot-gateway-sdk.git
+
+WORKDIR azure-iot-gateway-sdk/
+
+# Build node.dll for gateway's Node.js binding
+RUN cp tools/build_nodejs.sh . \
```
avranju: Actually we should be able to just directly run `tools/build_nodejs.sh`, i.e., just have the following:

```
RUN tools/build_nodejs.sh
```

tools/Dockerfile
```diff
@@ -1,42 +1,46 @@
...
+    && ./build_nodejs.sh \
+    && rm ./build_nodejs.sh
+
+ENV JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64 \
+    NODE_INCLUDE=/build_nodejs/dist/inc \
+    NODE_LIB=/build_nodejs/dist/lib
```
avranju: If we directly ran `tools/build_nodejs.sh` then the values for the `NODE_INCLUDE` and `NODE_LIB` variables will need to be prefixed with `/azure-iot-gateway-sdk`.

tools/Dockerfile
```diff
@@ -1,42 +1,46 @@
...
+
+# Trigger the population of the local package cache
+ENV NUGET_XMLDOC_MODE skip
+RUN mkdir /warmup \
+    && cd /warmup \
+    && dotnet new \
```
avranju: Interesting. If we are attempting to pre-populate the nuget cache, don't we need to do a `dotnet restore` as well?

tools/Dockerfile
```diff
@@ -1,42 +1,46 @@
...
+    git \
+    cmake \
+    maven \
+    valgrind \
+    pkg-config \
+    uuid-dev \
+    libcurl4-openssl-dev \
+    libssl-dev \
+    libglib2.0-dev \
+    dotnet-dev-1.0.0-preview2.1-003177
```
avranju: Are we still on the preview release of .NET core? Or should this be `dotnet-dev-1.0.1`?

damonbarry: This can be updated.

tools/Dockerfile
damonbarry: Yeah. I did it this way so that the build_nodejs folder doesn't get wiped every time I run `git clean -xdf .` (which I tend to do a lot).

tools/Dockerfile
damonbarry: I don't know...I got this from a tutorial. I don't think it's strictly necessary, just saves time when you run the container.

damonbarry:
Actually, the problem with the old Dockerfile was just pilot error. It still works--I'm able to build it an run it (didn't try to run a sample gateway or anything like that).

We could perhaps use this Dockerfile for ideas on how to update the existing one with support for .NET Core, for instance. But otherwise I'll just close this one for now.

avranju:
Yes, I just tried building it and it worked.

tools/Dockerfile
avranju: I see. Makes sense.

tools/Dockerfile
damonbarry: @avranju While I have the info on my terminal screen, here's what happens when you run `dotnet new` for the first time:

```
Damons-MBP:~ damonb$ dotnet new console -o hwapp

Welcome to .NET Core!
---------------------
Learn more about .NET Core @ https://aka.ms/dotnet-docs. Use dotnet --help to see available commands or go to https://aka.ms/dotnet-cli-docs.

Telemetry
--------------
The .NET Core tools collect usage data in order to improve your experience. The data is anonymous and does not include command-line arguments. The data is collected by Microsoft and shared with the community.
You can opt out of telemetry by setting a DOTNET_CLI_TELEMETRY_OPTOUT environment variable to 1 using your favorite shell.
You can read more about .NET Core tools telemetry @ https://aka.ms/dotnet-cli-telemetry.

Configuring...
-------------------
A command is running to initially populate your local package cache, to improve restore speed and enable offline access. This command will take up to a minute to complete and will only happen once.
Decompressing 100% 4546 ms
Expanding 100% 5597 ms
Getting ready...
Content generation time: 49.3261 ms
The template "Console Application" created successfully.
Damons-MBP:~ damonb$ 
```

...so it looks like `dotnet new` does an _initial, system-wide_ population of your local package cache, while `dotnet restore` will do a project-specific population/refresh.

avranju: Thanks! Makes sense.

PR #166, opened 2017-03-14
Example modules handle empty content better.

darobs:
When logger and Azure Functions receives a message with empty content, it reports content as empty, rather than considering that an error.

### DIFF
--- a/modules/azure_functions/src/azure_functions.c
+++ b/modules/azure_functions/src/azure_functions.c
@@ -304,7 +304,17 @@ static void AzureFunctions_Receive(MODULE_HANDLE moduleHandle, MESSAGE_HANDLE me
         else
         {
             /* Codes_SRS_AZUREFUNCTIONS_04_013: [ AzureFunctions_Receive shall base64 encode by calling Base64_Encode_Bytes, if it fails it shall fail and return. ] */
-            STRING_HANDLE contentAsJSON = Base64_Encode_Bytes(content->buffer, content->size);
+			STRING_HANDLE contentAsJSON;
+			if (content->buffer == NULL)
+			{
+				contentAsJSON = STRING_construct_n("", 0);
+			}
+			else
+			{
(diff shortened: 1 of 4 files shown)


modules/azure_functions/src/azure_functions.c
```diff
@@ -304,7 +304,15 @@ static void AzureFunctions_Receive(MODULE_HANDLE moduleHandle, MESSAGE_HANDLE me
...
+			STRING_HANDLE contentAsJSON;
+			if (content->buffer == NULL)
+			{
+				contentAsJSON = STRING_construct_n("", 0);
+			}
+			else
+			{
+				contentAsJSON = Base64_Encode_Bytes(content->buffer, content->size);
+			}
             if (contentAsJSON == NULL)
```
aribeironovaes: Suggest to put a line between this so it's easier to read (easy to mix with elseif if it's together.

modules/logger/src/logger.c
```diff
@@ -446,7 +446,22 @@ static void Logger_Receive(MODULE_HANDLE moduleHandle, MESSAGE_HANDLE messageHan
...
+								if (content->buffer == NULL)
+								{
+									contentAsJSON = STRING_construct_n("", 0);
+								}
+								else
+								{
+									contentAsJSON = Base64_Encode_Bytes(content->buffer, content->size);
+								}
+							}
                             if (contentAsJSON == NULL)
```
aribeironovaes: Same comment. New Line.

aribeironovaes:
:shipit:

modules/azure_functions/src/azure_functions.c
darobs: Put some whitespace and added a comment. for better clarity.

modules/logger/src/logger.c
darobs: Same fix: Put some whitespace and added a comment. for better clarity.

darobs:
Thank you, @aribeironovaes!

PR #169, opened 2017-03-18
Build gateway SDK for macOS

damonbarry:
These changes don't support the Node.js and .NET Core bindings yet, but I've been sitting on them awhile and I'd like to get them checked in. I'll do a separate PR when I add support for Node.js and .NET Core bindings.

### DIFF
--- a/core/tests/gateway_e2e/CMakeLists.txt
+++ b/core/tests/gateway_e2e/CMakeLists.txt
@@ -17,7 +17,7 @@ ${theseTestsName}.cpp
 #setting the dynamic_loader file based on OS that it is used
 if(WIN32)
     set(modules_c_file ./module_config_windows.c)
-elseif(LINUX)
+elseif(UNIX) # LINUX OR APPLE
     set(modules_c_file ./module_config_linux.c)
 endif()
 
--- a/core/tests/gwmessage_ut/gwmessage_ut.c
+++ b/core/tests/gwmessage_ut/gwmessage_ut.c
@@ -1147,7 +1147,7 @@ BEGIN_TEST_SUITE(gwmessage_ut)
         ///cleanup
     }
 
-    /*Tests_SRS_MESSAGE_02_020: [Otherwise, Message_Destroy shall decrement the internal ref count of the message.] 
+    /*Tests_SRS_MESSAGE_02_020: [Otherwise, Message_Destroy shall decrement the internal ref count of the message.]*/
     /*Tests_SRS_MESSAGE_02_021: [If the ref count is zero then the allocated resources are freed.]*/
     /*Tests_SRS_MESSAGE_17_002: [Message_Destroy shall destroy the CONSTMAP properties.]*/
     /*Tests_SRS_MESSAGE_17_005: [Message_Destroy shall destroy the CONSTBUFFER.]*/
--- a/jenkins/osx_c.sh
+++ b/jenkins/osx_c.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+set -e
+
+script_dir=$(cd "$(dirname "$0")" && pwd)
+build_root=$(cd "${script_dir}/.." && pwd)
+build_folder=$build_root"/build"
+
+CORES=$(grep -c ^processor /proc/cpuinfo 2>/dev/null || sysctl -n hw.ncpu)
+
+# Java binding
+pushd $build_root/bindings/java/gateway-java-binding
+mvn clean install
+popd
+
+rm -rf $build_folder
+mkdir -p $build_folder
+pushd $build_folder
+
+cmake \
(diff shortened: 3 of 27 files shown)


damonbarry:
I've confirmed the following samples work on my Mac:
- hello_world
- simulated_device_cloud_upload

I have _NOT_ tried these yet:
- azure_functions_sample
- native_module_host_sample

Finally, I tried these (briefly) but didn't get them to work. I'll need to investigate:
- java_sample
- proxy_sample

aribeironovaes:
Can you also test the Managed Gateway Sample? It would be great to check if they work on MAc as well! :D 


---
In reply to: [287583691](https://github.com/Azure/azure-iot-gateway-sdk/pull/169#issuecomment-287583691) [](ancestors = 287583691)

aribeironovaes:
Nevermind! Just saw the description that this change doesn't support .NEt Core Binding yet. 

---
In reply to: [288160517](https://github.com/Azure/azure-iot-gateway-sdk/pull/169#issuecomment-288160517) [](ancestors = 288160517,287583691)

jenkins/osx_c.sh
```diff
@@ -0,0 +1,35 @@
...
+
+CORES=$(grep -c ^processor /proc/cpuinfo 2>/dev/null || sysctl -n hw.ncpu)
+
+# Java binding
+pushd $build_root/bindings/java/gateway-java-binding
+mvn clean install
+popd
+
+rm -rf $build_folder
+mkdir -p $build_folder
```
aribeironovaes: What happens if these command fails? Do you have error checking?

aribeironovaes:
:shipit:

jenkins/osx_c.sh
damonbarry: Yes, `set -e` causes the script to exit if any command fails. `set -e` has it's caveats--it's not a panacea. But I think it works fine for this script so far (e.g., IIRC the may be problems for commands in `if` statements, but we don't have any here yet).

core/tests/gateway_e2e/CMakeLists.txt
```diff
@@ -19,6 +19,8 @@ if(WIN32)
     set(modules_c_file ./module_config_windows.c)
 elseif(LINUX)
     set(modules_c_file ./module_config_linux.c)
+elseif(APPLE)
+    set(modules_c_file ./module_config_osx.c)
```
avranju: If it's a new file maybe call it *module_config_macos.c*? 😛

core/tests/gwmessage_ut/gwmessage_ut.c
```diff
@@ -1147,7 +1147,7 @@ BEGIN_TEST_SUITE(gwmessage_ut)
         ///cleanup
     }
 
-    /*Tests_SRS_MESSAGE_02_020: [Otherwise, Message_Destroy shall decrement the internal ref count of the message.] 
+    /*Tests_SRS_MESSAGE_02_020: [Otherwise, Message_Destroy shall decrement the internal ref count of the message.]*/
```
avranju: Whoa. Good eye!

modules/azure_functions/src/azure_functions.c
```diff
@@ -393,7 +393,8 @@ static void AzureFunctions_Receive(MODULE_HANDLE moduleHandle, MESSAGE_HANDLE me
                                             {
 
                                                 //Add here the content on the body.
-                                                BUFFER_HANDLE postContent = BUFFER_create(STRING_c_str(jsonToBeAppended), STRING_length(jsonToBeAppended));
+                                                size_t appendLength = STRING_length(jsonToBeAppended);
```
avranju: Just curious, why is it better if this length is first stored in a local?

modules/identitymap/tests/idmap_ut/idmap_ut.cpp
```diff
@@ -431,7 +431,7 @@ TYPED_MOCK_CLASS(CIdentitymapMocks, CGlobalMock)
     MOCK_STATIC_METHOD_2(, void*, VECTOR_element, const VECTOR_HANDLE, handle, size_t, index)
         currentVectorElement_call++;
         void * result1;
-        if (currentVectorElement_call = whenShallVectorElement_fail)
+        if (currentVectorElement_call == whenShallVectorElement_fail)
```
avranju: Wowza! Nice catch!

core/tests/gwmessage_ut/gwmessage_ut.c
damonbarry: clang ;)

modules/azure_functions/src/azure_functions.c
damonbarry: The important thing was that we couldn't have more than one function call in the arguments to `BUFFER_create` because of platform differences RE the order in which arguments are popped off the stack. Unit tests that use extensive mocking + EXPECTED_CALL will pass on one platform and fail on another. ☹️

modules/identitymap/tests/idmap_ut/idmap_ut.cpp
damonbarry: clang

PR #170, opened 2017-03-19
Update README.md

DBojsen:
Adding Proficy Historian module


damonbarry:
@DBojsen This is great!

Looking at our [Contribution Guidelines](https://github.com/Azure/azure-iot-gateway-sdk/blob/master/Contributing.md#get-your-module-featured-in-our-readme), one of the requirements for getting an OSS module featured is that you provide some level of automated tests, so users can validate that the module does what it says it does. Can you add some tests for this module?

At a basic level, you could add a single end-to-end test that passes if it can connect to a Proficy Historian server, receive some data from it, and publish the data to the gateway's Broker. Of course the user would have to supply the Proficy Historian server connection info (unless there's a way to mock the historian API).

You might have noticed that one of the modules we're already featuring is missing tests. 🤕 That's a problem we hope to correct.

DBojsen:
Hi Damon, Angelo

I have added unit tests as a separate project.

BR David

PR #175, opened 2017-03-22
Update README.md

606simps:
Add SQLite module to the Featured Modules table.
Update Modbus module version.


aribeironovaes:
@606simps 
Thanks a lot for your contribution! 

Looking at our Contribution Guidelines, one of the requirements for getting an OSS module featured is that you provide some level of automated tests, so users can validate that the module does what it says it does. Can you add some tests for this module?

It can be a simple end2end test, where the user can configure what they need in order to run your module and run this test to check if it's working with their module.

You might have noticed that one of the modules we're already featuring is missing tests. 🤕 That's a problem we hope to correct.

606simps:
Thanks for the input. I've updated the SQLite module with unit tests. Let me know if anything else is yet missing.

PR #179, opened 2017-03-22
Send start message when oop module is recreated

ancaantochi:


proxy/outprocess/src/module_loaders/outprocess_module.c
```diff
@@ -395,6 +395,8 @@ int outprocessControlThread(void *param)
 				}
 				else
 				{
+                    /*Codes_SRS_OUTPROCESS_MODULE_24_061: [ Once the control channel has been restarted and Create Message was sent, it shall send a Start Message to the module host. ]*/
+                    send_start_message(handleData);
```
aribeironovaes: Hi Anca, Shall we test here if Gateway is started before sending start message? Think about this scenario: 1-User call Gateway_CReate (No CreateFromJson);2-User Add Modules (Without Start);3-User Stop Gateway;4-User Add an OUt of Process Module). On this scenario, wouldn't we call start even if the gateway is not started?

ancaantochi: outprocessControlThread is created in Outprocess_Start, that means that the gateway was started when it can reach this code to recreate the out of process module.

aribeironovaes: Cool. Thanks Anca.

PR #183, opened 2017-03-24
Support Visual Studio 2017 builds

damonbarry:
Had to fix some things in our dependencies, then do a submodule dance, and finally update the build script.

The build script uses your Visual Studio environment to determine what to build. If you're in a "Developer Command Prompt for VS2015", then you'll build for VS 2015. If you're in a "Developer Command Prompt for 2017", then you'll build for 2017. If it's an environment that doesn't define VisualStudioVersion (e.g. vanilla command prompt), then you'll get an error (at least for now).

### DIFF
--- a/tools/build.cmd
+++ b/tools/build.cmd
@@ -157,6 +157,21 @@ rem ----------------------------------------------------------------------------
 rem this is setting the cmake path in a quoted way
 set "cmake-root=%build-root%\build"
 
+rem Figure out which CMake generator to use
+if "%VisualStudioVersion%" == "14.0" (
+    set "cmake-generator=Visual Studio 14 2015"
+) else if "%VisualStudioVersion%" == "15.0" (
+    set "cmake-generator=Visual Studio 15 2017"
+)
+if NOT DEFINED cmake-generator (
+    echo ERROR: VisualStudioVersion not defined, or not one of ^(14.0, 15.0^)
+    exit /b 1
+)
+if "%build-platform%" == "x64" (
+    set "cmake-generator=%cmake-generator% Win64"
+)
+echo **Building for "%cmake-generator%"
+
 echo Cleaning up build artifacts...
 rmdir /s/q %cmake-root%
 if not !ERRORLEVEL!==0 exit /b !ERRORLEVEL!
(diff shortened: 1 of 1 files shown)


damonbarry:
@Azure/iot-gateway-devs

tools/build.cmd
```diff
@@ -157,6 +157,21 @@ rem ----------------------------------------------------------------------------
...
+    set "cmake-generator=Visual Studio 14 2015"
+) else if "%VisualStudioVersion%" == "15.0" (
+    set "cmake-generator=Visual Studio 15 2017"
+)
+if NOT DEFINED cmake-generator (
+    echo ERROR: VisualStudioVersion not defined, or not one of ^(14.0, 15.0^)
+    exit /b 1
+)
+if "%build-platform%" == "x64" (
+    set "cmake-generator=!cmake-generator! Win64"
```
avranju: Didn't know batch files supported variable expansion using `!` symbols! Is the usual `%` syntax not permissible here?

tools/build.cmd
```diff
@@ -157,6 +157,21 @@ rem ----------------------------------------------------------------------------
...
+)
+echo **Building for "%cmake-generator%"
```
avranju: Nice!

tools/build.cmd
damonbarry: Actually, the ! syntax isn't really necessary here; I'll change it.

You need to use the ! syntax when you **set** and then **use** a variable **inside an `if` statement**. E.g.
```
@setlocal
@echo off
set hello=world
echo %hello%
if "%hello%" == "world" (
    set "hello=cruel %hello%"
    echo %hello%
)
echo %hello%
endlocal
```
...gives this output:
```
world
world
cruel world
```
Oops! The 2nd line should say "cruel world", but since the entire if-block was evaluated before it was executed, `%hello%` resolved to its pre-if-block value. But change it to this:
```
@setlocal EnableDelayedExpansion
@echo off
set hello=world
echo %hello%
if "%hello%" == "world" (
    set "hello=cruel %hello%"
    echo !hello!
)
echo %hello%
endlocal
```
...and you get the expected output:
```
world
cruel world
cruel world
```
Expansion of the variable inside the !s is delayed until the line is actually executed, so it gets the right value.

avranju: Wow. Thanks for the detailed explanation. Sounds like an afterthought of a feature! As part of the Node.js get-off-the-fork work I am switching the Node.js build script to use PowerShell because of batch file parameter parsing weirdness. There will still be a `build_nodejs.cmd` file but that's just a small front-end for the PS script. Coming from batch files, PS now feels like a breath of fresh air!

PR #198, opened 2017-04-04
Activation Type: Launch

zafields:
NOT ready for merge, tests to follow.


proxy/outprocess/src/module_loaders/outprocess_loader.c
```diff
@@ -24,344 +31,616 @@ DEFINE_ENUM_STRINGS(OUTPROCESS_LOADER_ACTIVATION_TYPE, OUTPROCESS_LOADER_ACTIVAT
...
 typedef struct OUTPROCESS_MODULE_HANDLE_DATA_TAG
 {
-	const MODULE_API* api;
+    const MODULE_API* api;
+
 } OUTPROCESS_MODULE_HANDLE_DATA;
 
+static VECTOR_HANDLE uv_processes = NULL;
+static THREAD_HANDLE uv_thread = NULL;
+static tickcounter_ms_t uv_process_grace_period_ms = 0;
```
darobs: can these be put somewhere, instead of static variables?  I'll make a suggestion when I get more into the code.

zafields: No, the API's related to this data fall outside the ModuleLoader API, and therefore the API's and their respective data must be static in order to be called from the gateway.

zafields: In an object oriented world, this would be a class level operation.

core/CMakeLists.txt
```diff
@@ -32,6 +32,21 @@ elseif(UNIX) # LINUX or APPLE
     set(dynamic_library_c_file ./adapters/dynamic_library_linux.c ./adapters/gb_library_linux.c )
 endif()
 
+# If using the appropriate build script based on OS that it is used
+if (${enable_native_remote_modules} OR ${enable_java_remote_modules})
```
darobs: What is the decision behind building libuv versus installing it? Should we check to see if the package is installed on the system before building it?

zafields: Yeah, we should. I have modified the cmake just enough to make Windows work and I am soliciting help.

proxy/outprocess/src/module_loaders/outprocess_loader.c
```diff
@@ -20,348 +27,699 @@ DEFINE_ENUM_STRINGS(OUTPROCESS_LOADER_ACTIVATION_TYPE, OUTPROCESS_LOADER_ACTIVAT
...
+        {
+            // Wait for child to clean up
+            tickcounter_ms_t now = started_waiting;
+            /* Codes_SRS_OUTPROCESS_LOADER_27_058: [ `OutprocessLoader_JoinChildProcesses` shall await the grace period in 100ms increments, by calling `void ThreadAPI_Sleep(unsigned int milliseconds)` passing 100 for `milliseconds`. ] */
+            for (timed_out = true; (now - started_waiting) < uv_process_grace_period_ms; ThreadAPI_Sleep(GRACE_AWAIT_DELAY_MS))
+            {
+                /* Codes_SRS_OUTPROCESS_LOADER_27_055: [ While awaiting the grace period, `OutprocessLoader_JoinChildProcesses` shall mark the current time, by calling `int tickcounter_get_current_ms(TICK_COUNTER_HANDLE tick_counter, tickcounter_ms_t * current_ms)` using the handle called by the previous call to `tickcounter_create`. ] */
+                if (tickcounter_get_current_ms(ticks, &now))
+                {
+                    LogError("failed to sample tickcounter");
```
darobs: I am assuming that a failure here would want the same behavior as timing out...  Because if we hit this condition, we're going to kill all the spawned processes.  Am I correct?

samples/proxy_sample/src/remote_main.c
```diff
@@ -56,6 +56,16 @@ const MODULE_API * pRemoteModuleApi = (const MODULE_API *)&remoteModuleApi;
 int main(int argc, char** argv)
 {
     REMOTE_MODULE_HANDLE remote_module;
+
+    // Log received parameters to file
+    FILE * param_log = fopen("param.txt", "w+");
```
damonbarry: Did you mean to check in this param.txt stuff? Seems unusual...

aribeironovaes:
Suggestion: SQUASH these 11 commits. Comments like: tweeks for LInux or Tweeks for Windows, please remove. :)

aribeironovaes:
:shipit:

PR #204, opened 2017-04-11
Updated Dockerfile to add support for building DotNet bindings and te…

mrohera:
…sts. Additionally used a slimmer image and removed the need to obtain the build_nodejs.sh before building the container image.

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -1,42 +1,93 @@
-FROM ubuntu:14.04
+FROM debian:jessie-slim
 
-RUN apt-get clean && \
-    apt-get update && \
-    apt-get install -y \
-        software-properties-common && \
-    apt-add-repository ppa:webupd8team/java && \
-    echo oracle-java8-installer shared/accepted-oracle-license-v1-1 select true | sudo /usr/bin/debconf-set-selections && \
-    apt-get update && \
-    apt-get install -y \
-        oracle-java8-installer \
-        oracle-java8-set-default \
-        maven \
-        python2.7 \
-        curl \
-        build-essential \
-        libcurl4-openssl-dev \
-        git \
-        cmake \
-        libssl-dev \
-        uuid-dev \
-        valgrind \
-        libglib2.0-dev
+########### Install .NET, Java, Node and Gateway dependencies ###########
+    RUN echo "===> Install Gateway Dependencies..." \
+    ########### Install Java ########### \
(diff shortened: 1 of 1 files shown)


tools/Dockerfile
```diff
@@ -1,42 +1,95 @@
...
+    ENV NODE_LIB /build_nodejs/dist/lib
+
+########### Install Gateway ###########
+    ENV GATEWAY_REPO https://github.com/Azure/azure-iot-gateway-sdk
+    ENV COMMIT_ID master
+
+    # start command
+    ENTRYPOINT  rm -rf /gateway \
+                && git clone ${GATEWAY_REPO} /gateway \
+                && git -C /gateway checkout ${COMMIT_ID} \
```
damonbarry: `master` is the default, so this is a no-op

tools/Dockerfile
```diff
@@ -1,42 +1,95 @@
...
+                && git -C /gateway submodule update --init \
```
damonbarry: This step isn't necessary. The build process will init submodules as needed.

tools/Dockerfile
```diff
@@ -1,42 +1,95 @@
...
-
-ENTRYPOINT rm -rf /gateway && \
-           git clone ${GATEWAY_REPO} /gateway && \
-           git -C /gateway checkout ${COMMIT_ID} && \
-           git -C /gateway submodule update --init && \
-           /gateway/tools/build.sh --run-unittests -rv --enable-java-binding --enable-nodejs-binding && \
-           bash
+FROM debian:jessie-slim
+
+########### Install .NET, Java, Node and Gateway dependencies ###########
```
damonbarry: Consider making each of these major steps (the sections set apart by a `###########` comment) its own `RUN` statement so that if one fails, `docker build` doesn't have to start over from the beginning.

tools/Dockerfile
damonbarry: When this script was written we had a `develop` branch, but that's no longer the case so this isn't needed.

tools/Dockerfile
mrohera: If you look at https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run specifically for apt-get and other "apt-" variants, the guideline is to keep together. My attempt to use hashes was to break up the deps and categorize by technology.

tools/Dockerfile
mrohera: Makes sense to cleanup then. Thanks for letting me know.

tools/Dockerfile
mrohera: Yes that is correct, I can clean up these lines as well.

PR #206, opened 2017-04-11
Improve channel identifiers

damonbarry:
1. Don't URL-encode control/message channel identifiers. The valid character sets for identifiers varies by platform, and on no platform does the valid character set conform to the RFCs for URL encoding. We'll let the underlying platform technologies (used by nanomsg) validate the identifier.

2. Don't add a ".ipc" suffix to the channel identifiers. On Posix-based platforms the identifier is actually a file path. If we change the file path in any way then it won't be where the user might expect it.


damonbarry:
@Azure/iot-gateway-devs

PR #211, opened 2017-04-14
Silence annoying VECTOR errors when using Java modules

damonbarry:
@ancaantochi @Azure/iot-gateway-devs


bindings/java/tests/host_ut/host_ut.c
```diff
@@ -1525,8 +1523,6 @@ TEST_FUNCTION(JavaModuleHost_Create_initializes_JavaVMInitArgs_structure_no_addi
     STRICT_EXPECTED_CALL(JavaModuleHostManager_Create(&config2));
 
     STRICT_EXPECTED_CALL(VECTOR_create(sizeof(STRING_HANDLE)));
```
damonbarry: If the vector is non-NULL *and* non-empty, it should take the same code path. And I've verified that there is a test that passes in an non-empty vector. So we do have coverage in this case.

PR #222, opened 2017-04-19
Performance test code

darobs:


core/tests/CMakeLists.txt
```diff
@@ -34,5 +35,6 @@ endif()
 
 if(${run_e2e_tests})
     add_subdirectory(gateway_e2e)
+    add_subdirectory(performance_e2e)
```
aribeironovaes: >add_subdirectory(performance_e2e) [](start = 4, length = 33)

I think we should separate the performance_e2e from the e2e test, don't you think? have it's own flag run_performance_tests

core/tests/performance_e2e/src/metrics.cpp
```diff
@@ -0,0 +1,279 @@
...
+#include <string>
+#include <vector>
+#include <map>
+#include <utility>
+#include <algorithm>
+#include <random>
+#include <exception>
+
+#include <parson.h>
+
```
aribeironovaes: Do we need all of these?

core/tests/performance_e2e/src/metrics.cpp
```diff
@@ -0,0 +1,279 @@
...
+                << "Devices Discovered: " << module->per_device_metrics->size() << std::endl;
+            for (PerDeviceMap::iterator d = module->per_device_metrics->begin();
+                d != module->per_device_metrics->end();
+                d++)
+            {
+                std::cout
+                    << "Device: " << (*d).first << std::endl
+                    << "Message count: " << (*d).second.messages_received << std::endl
+                    << "Out of Sequence Count: " << (*d).second.out_of_sequence_messages << std::endl
+                    << "Messages Lost: " << (*d).second.messages_lost << std::endl;
```
aribeironovaes: Shall we save this in a file instead of printing in a screen?

core/tests/performance_e2e/src/metrics.cpp
```diff
@@ -0,0 +1,279 @@
...
+        METRICS_MODULE_HANDLE * module = (METRICS_MODULE_HANDLE *)moduleHandle;
+        if (module->started)
+        {
+            HrTime destroy_time = std::chrono::time_point_cast<MicroSeconds>(HrClock::now());
+            MicroSeconds duration = destroy_time - module->start_time;
+            std::cout
+                << "Module Metrics:" << std::endl
+                << "---------------" << std::endl
+                << "Duration: " << duration.count() / 1000 << std::endl
+                << "Messages received: " << module->all_messages_received << std::endl
```
aribeironovaes: Shall we put the metric? (s, ms or us?)

core/tests/performance_e2e/src/metrics.cpp
darobs: Nope, I narrowed it down.

core/tests/performance_e2e/src/metrics.cpp
darobs: Yes, I should.  Done.

core/tests/CMakeLists.txt
darobs: I'm going to leave it as e2e for now - if we get more tests like this, then it might make more sense to break them into their own category.

core/tests/performance_e2e/src/metrics.cpp
darobs: I think I'm going to keep it printing to the screen, because the next step is to add it as a test that runs as part of the build.

PR #229, opened 2017-04-24
Perf test  e2e run in build

darobs:


core/tests/performance_e2e/src/module_config_windows.c
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "gateway.h"
+
+const char* simulator_module_path_string = "Debug\\simulator.dll";
```
aribeironovaes: >Debug\ [](start = 44, length = 6)

Why hardcoded to Debug? What happens if we build Release?

aribeironovaes:
:shipit:

PR #236, opened 2017-04-27
Bump versions for 2017-04-27 release

damonbarry:
Code changes are limited to the core (which includes loaders, the native part of bindings, etc) and the .NET Core assembly (changed the target from netstandard16 to netstandard13).

### DIFF
--- a/tools/release/bump_version/versions.json
+++ b/tools/release/bump_version/versions.json
@@ -1,12 +1,12 @@
 {
     "gateway": 
     {
-        "core" : "1.0.4"
+        "core" : "1.0.5"
     },
     "bindings" :
     {
         "dotnet" : "1.0.2",
-        "dotnetcore": "1.0.2",
+        "dotnetcore": "1.0.3",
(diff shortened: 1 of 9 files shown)


damonbarry:
@Azure/iot-gateway-devs

tools/release/bump_version/versions.json
```diff
@@ -1,12 +1,12 @@
...
     {
-        "core" : "1.0.4"
+        "core" : "1.0.5"
     },
     "bindings" :
     {
         "dotnet" : "1.0.2",
-        "dotnetcore": "1.0.2",
+        "dotnetcore": "1.0.3",
         "java" : "1.1.0"
```
aribeironovaes: Didn't we fix a bug on Java binding? Shall we increase the minor version?

damonbarry: We didn't touch any .java files, so we don't need to bump versions inside the JARs. We made some changes to C code like the module loader, but that is considered part of "core" for now...

PR #240, opened 2017-05-01
Update README for LibUV prerequisites

zafields:


doc/devbox_setup.md
```diff
@@ -21,6 +21,10 @@ This section shows you how to set up a development environment for the Azure IoT
 
     > Note: Using the .msi is the easiest option when installing on Windows. Add CMake to the PATH for at least the current user when prompted to do so by the installer.
 
+1. Install [Python 2.7](https://www.python.org/downloads/release/python-27)
```
avranju: Don't we need to add this for the Linux setup steps too?

PR #251, opened 2017-05-09
Update Dockerfile to build libuv

damonbarry:
- Get a newer version of CMake from `jessie-backports`
- Install necessary dependencies (automake, libtool)
- Rename gateway SDK to IoT Edge
- Whitespace and other superfluous changes...not because they're more correct, but because I couldn't help myself.

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -1,93 +1,94 @@
 FROM debian:jessie-slim
 
-########### Install .NET, Java, Node and Gateway dependencies ###########
-    RUN echo "===> Install Gateway Dependencies..." \
-    ########### Install Java ########### \
-        && echo "deb http://ppa.launchpad.net/webupd8team/java/ubuntu trusty main" | tee /etc/apt/sources.list.d/webupd8team-java.list \
-        && echo "deb-src http://ppa.launchpad.net/webupd8team/java/ubuntu trusty main" | tee -a /etc/apt/sources.list.d/webupd8team-java.list \
-        && apt-key adv --keyserver keyserver.ubuntu.com --recv-keys EEA14886 \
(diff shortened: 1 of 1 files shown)


tools/Dockerfile
```diff
@@ -1,93 +1,94 @@
...
+        libuuid1 \
+        zlib1g
 
-    RUN curl -SL $DOTNET_SDK_DOWNLOAD_URL --output dotnet.tar.gz \
-        && mkdir -p /usr/share/dotnet \
-        && tar -zxf dotnet.tar.gz -C /usr/share/dotnet \
-        && rm dotnet.tar.gz \
-        && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
+#### Install Libuv dependencies
+RUN apt-get install -y automake libtool \
```
zafields: RUN creates a "commit" in the docker image. Do these two packages need to be in a separate commit, or is this just an unintended consequence of the comment?

A snippet from the [documentation](https://docs.docker.com/engine/reference/builder/#run):

> The RUN instruction will execute any commands in a new layer on top of the current image and commit the results. The resulting committed image will be used for the next step in the Dockerfile.
Layering RUN instructions and generating commits conforms to the core concepts of Docker where commits are cheap and containers can be created from any point in an image’s history, much like source control.

damonbarry: Yeah, there's always a balance between too many or too few steps packed into a single commit. In the end, though, I just don't think it matters all that much for this developer image. I went for clarity in the Dockerfile over efficiency (# commits). Before, we were doing some acrobatics to pack things into fewer commits, and it just made the Dockerfile that much harder to read.

zafields: word.

PR #255, opened 2017-05-12
fix headers for github markdown

arehman199:
A space is required between the # signs and the headings text.

Signed-off-by: Abdur Rehman <abdur_rehman@mentor.com>


damonbarry:
No activity...closing.

PR #266, opened 2017-05-15
"Updated hw sample json to run from root"

jonbgallant:


zafields:
I think this breaks consistency with our other samples, and we should also ensure other examples don't have a dependency on this sample.

damonbarry:
I updated the hello world README file to instruct users to run the sample from the `build/` directory. Closing this PR.

PR #283, opened 2017-05-25
Support Node.js apps as out-of-proc modules

damonbarry:
I'll need to update the Jenkins job(s) to lint and run the unit tests for this feature.

### DIFF
--- a/proxy/gateway/nodejs/index.js
+++ b/proxy/gateway/nodejs/index.js
@@ -1,0 +1,69 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+let ChannelFactory = require('./lib/channel_factory.js');
+let AttachedError = require('./lib/exceptions.js').AttachedError;
+
+class ProxyGateway {
+  constructor(module, channelFactory) {
+    if (module === undefined) throw new TypeError("Missing 'module' argument");
+    this.module = module;
+    this.channelFactory = channelFactory || ChannelFactory;
+    this.controlChannel = null;
+    this.messageChannel = null;
+  }
+
+  _disconnectMessageChannel() {
+    if (this.messageChannel) {
+      this.messageChannel.disconnect();
+      this.messageChannel = null;
+    }
+  }
+
+  onMessage(msg) {
+    this.module.receive(msg);
+  }
+
+  onCreate(messageChannelId, args) {
+    if (this.messageChannel) this.onDestroy();
+    this.messageChannel = this.channelFactory.createMessageChannel();
+    this.messageChannel.connect(messageChannelId);
+    this.messageChannel.on('message', this.onMessage.bind(this));
+    let broker = {
+      publish: (msg) => this.messageChannel.send(msg)
+    };
+    let result = this.module.create(broker, args);
+    this.controlChannel.send('create', result);
+  }
+
+  onStart() {
+    if (typeof(this.module.start) === 'function') {
+      this.module.start();
+    }
+  }
+
+  onDestroy() {
+    this.module.destroy();
+    this._disconnectMessageChannel();
+  }
+
+  attach(controlChannelId) {
+    if (this.controlChannel) throw new AttachedError('Already attached');
+    this.controlChannel = this.channelFactory.createControlChannel();
+    this.controlChannel.connect(controlChannelId);
+    this.controlChannel.on('create', this.onCreate.bind(this));
+    this.controlChannel.on('start', this.onStart.bind(this));
+    this.controlChannel.on('destroy', this.onDestroy.bind(this));
+  }
+
+  detach() {
+    this.controlChannel.send('detach');
+    this._disconnectMessageChannel();
+    this.controlChannel.disconnect();
+    this.controlChannel = null;
+  }
+}
+
+module.exports = ProxyGateway;
--- a/proxy/gateway/nodejs/lib/channel.js
+++ b/proxy/gateway/nodejs/lib/channel.js
@@ -1,0 +1,38 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+const channelUriPrefix = require('./constants.js').channelUriPrefix;
+
+let EventEmitter = require('events').EventEmitter;
+let nano = require('nanomsg');
+let fs = require('fs');
+
+class Channel extends EventEmitter {
+  constructor() {
+    super();
+    this.socket = nano.socket('pair');
+    this.socket.on('data', data => this.emit('data', data));
+  }
+
+  connect(id) {
+    let addr = (typeof(id) === 'string') ? channelUriPrefix + id : id.uri;
+    try {
+      fs.unlinkSync(addr.replace(channelUriPrefix, ''));
+    }
+    catch (e) { }
+    this.socket.bind(addr);
+  }
+
+  send(msg) {
+    this.socket.send(msg);
+  }
+
+  disconnect() {
+    this.socket.linger();
+    this.socket.close();
+  }
+}
+
+module.exports = Channel;
--- a/proxy/gateway/nodejs/lib/channel_factory.js
+++ b/proxy/gateway/nodejs/lib/channel_factory.js
@@ -1,0 +1,19 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+let MessageChannel = require('./message_channel.js');
+let ControlChannel = require('./control_channel.js');
+
+class ChannelFactory {
+  static createMessageChannel() {
+    return new MessageChannel();
+  }
+
+  static createControlChannel() {
+    return new ControlChannel();
+  }
+}
+
+module.exports = ChannelFactory;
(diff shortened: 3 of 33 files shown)


damonbarry:
BTW, here are the "requirements" 😜 (aka unit test output):
```
  Channel
    ✓ can send a message (109ms)
    ✓ can receive a message (103ms)
    ✓ acts like an EventEmitter

  Codec
    #encodeControlMessage
      ✓ encodes a control message
    #decodeControlMessage
      ✓ decodes a control message
      ✓ throws DecodeError if the message doesn't have header bytes
      ✓ throws RangeError if the message is too short
    #decodeMessageChannelUri
      ✓ decodes a message channel URI
      ✓ returns the number of bytes read from the buffer
      ✓ reads the right number of bytes for the URI string
      ✓ throws RangeError if the buffer is too short
      ✓ optionally reads the buffer from the given offset
    #decodeCreateMessage
      ✓ decodes a 'create' message
      ✓ throws RangeError if the module arguments are too short
      ✓ reads the right number of bytes for the module args
    #encodeCreateReply
      ✓ encodes a successful 'create' reply message
      ✓ encodes a failed 'create' reply message
    #encodeModuleMessageProperties
      ✓ returns an empty buffer when given an empty object
      ✓ throws when a property value is null
      ✓ encodes module message properties
    #decodeModuleMessageProperties
      ✓ returns an empty object when given an empty buffer
      ✓ returns an empty object when asked to decode zero properties
      ✓ throws when a property key is not null-terminated
      ✓ throws when a property value is not null-terminated
      ✓ throws when a property has no value
      ✓ only decodes `count` properties
      ✓ decodes multiple properties
      ✓ returns the offset of the first byte after the properties
    #encodeModuleMessage
      ✓ throws if `properties` is null
      ✓ throws if `content` is not a Buffer
      ✓ throws if `content` is not a Buffer
      ✓ throws if `content` is not a Buffer
      ✓ throws if `content` is not a Buffer
      ✓ encodes an empty message from an empty object
      ✓ encodes a module message
    #decodeModuleMessage
      ✓ throws when the message doesn't start with header bytes 0xA160
      ✓ decodes a message

  ControlChannel
    ✓ can receive a 'create' message (107ms)
    ✓ can send a 'create' reply message (102ms)
    ✓ can receive a 'start' message (104ms)
    ✓ can receive a 'destroy' message (102ms)
    ✓ can send a 'detach' message (105ms)
    #send
      ✓ throws TypeError if 'op' is not 'create' or 'detach'

  MessageChannel
    ✓ can receive a message (103ms)
    ✓ can send a message (103ms)

  ProxyGateway
    #constructor
      ✓ requires a 'module' argument
    #attach
      ✓ connects the control channel
      ✓ sets up a listener for the 'create' event
      ✓ sets up a listener for the 'start' event
      ✓ sets up a listener for the 'destroy' event
      ✓ throws if already attached
      ✓ doesn't throw if detached
    #detach
      ✓ notifies the control channel that the application is detaching
      ✓ disconnects the message channel if needed
      ✓ disconnects the control channel
    #onCreate
      ✓ is called when the control channel emits 'create'
      ✓ connects the message channel with the given id
      ✓ sets up a listener for the 'message' event
      ✓ creates the module
      ✓ notifies the control channel when the module is created
      ✓ notifies the control channel when module creation fails
    #onMessage
      ✓ is called when the message channel emits 'message'
      ✓ gives the message to the module
    #onStart
      ✓ is called when the control channel emits 'start'
      ✓ doesn't call start if the module doesn't implement it
      ✓ starts the module
    #onDestroy
      ✓ is called when the control channle emits 'destroy'
      ✓ destroys the module
      ✓ disconnects the message channel

  69 passing (992ms)
```

proxy/gateway/nodejs/lib/channel_factory.js
```diff
@@ -0,0 +1,19 @@
...
+
+class ChannelFactory {
+  createMessageChannel() {
+    return new MessageChannel();
+  }
+
+  createControlChannel() {
+    return new ControlChannel();
+  }
+}
```
avranju: Make these 2 methods static?

proxy/gateway/nodejs/lib/codec.js
```diff
@@ -0,0 +1,190 @@
...
+
+function _uriTypeToString(type) {
+    if (type === nano._bindings.NN_PAIR) return 'pair';
+    return '';
+}
+
+function encodeControlMessage(obj) {
+  let buffer = Buffer.alloc(4);
+  buffer.writeUInt8(0xA1, 0);
+  buffer.writeUInt8(0x6C, 1);
```
avranju: Should we use constants for these magic numbers? Here and on line 40.

proxy/gateway/nodejs/lib/codec.js
```diff
@@ -0,0 +1,190 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
```
avranju: Is there a spec that documents this protocol? Maybe a comment with a link to the spec would be nice.

proxy/gateway/nodejs/lib/control_channel.js
```diff
@@ -0,0 +1,43 @@
...
+      else if (msg.type === codec.controlMessageTypes.start) {
+        this.emit('start');
+      }
+      else if (msg.type === codec.controlMessageTypes.destroy) {
+        this.emit('destroy');
+      }
+    });
+  }
+
+  send(op, arg) {
```
avranju: Nit: The signature of `ControlChannel.send` is different from `Channel.send`. Should `op` and `arg` be merged into a single parameter?

proxy/gateway/nodejs/index.js
```diff
@@ -0,0 +1,68 @@
...
+      this.messageChannel = null;
+    }
+  }
+
+  onMessage(msg) {
+    this.module.receive(msg);
+  }
+
+  onCreate(messageChannelId, args) {
+    this.messageChannel = this.channelFactory.createMessageChannel();
```
avranju: Can `this.messageChannel` be non-`null`? I suppose the question is - is it possible for `onCreate` to get called multiple times on the same instance?

proxy/gateway/nodejs/lib/channel.js
```diff
@@ -0,0 +1,36 @@
...
+
+class Channel extends EventEmitter {
+  constructor() {
+    super();
+    this.socket = nano.socket('pair');
+    this.socket.on('data', data => this.emit('data', data));
+  }
+
+  connect(id) {
+    let addr = (typeof(id) === 'string') ? 'ipc://' + id : id.uri;
```
damonbarry: Added a commit to throw when the message channel URI doesn't start with `ipc://`

proxy/gateway/nodejs/lib/channel.js
```diff
@@ -0,0 +1,36 @@
...
+    try {
+      fs.unlinkSync(addr.replace('ipc://', ''));
```
damonbarry: Exactly...nanomsg uses Unix domain sockets on posix systems, so the URI is a path to the file. And for whatever reason they don't clean up the file (even though I found GitHub issues where they added code to do this, ultimately it doesn't appear to be happening) so we're doing it here.

Since Windows doesn't use files, I wrapped it in a `try` so that we'll try to delete a file that doesn't exist, we'll swallow the error, and we'll continue. Yes we can consider some other platform switching mechanism. I'll talk to you about this...

proxy/gateway/nodejs/test/test_endpoint.js
```diff
@@ -0,0 +1,29 @@
...
+    this.socket = nano.socket('pair');
+    this.socket.connect('ipc://' + id);
+    this._received = new Promise((resolve) => {
+      this.socket.on('data', data => resolve(data));
+    });
+  }
+
+  send(msg) { this.socket.send(msg); }
+
+  receive() { return this._received; }
```
damonbarry: Yes, that's how I'm using it. But this is a test class, only used in tests to act as the _other_ side of a nanomsg session. Because I implement the receive function as a Promise, it allows me to do stuff like this:
```
    return listener.receive()
      .should.eventually.deep.equal(Buffer.from('hello'));
```

proxy/gateway/nodejs/lib/channel_factory.js
damonbarry: Cool, never used static before. Done!

proxy/gateway/nodejs/lib/codec.js
damonbarry: Just pushed some commits that introduce a new document, `proxy/message_format.md`. And I added a comment to `codec.js` to link to this document.

proxy/gateway/nodejs/lib/codec.js
damonbarry: Done.

proxy/gateway/nodejs/lib/control_channel.js
damonbarry: Agreed. `send` isn't really used polymorphically (OTOH `connect` and `disconnect` are), so I shouldn't try to make it look like it is. I should probably name the base `send` something else. But I think I'll look the other way, at least for now 😉

proxy/gateway/nodejs/test/test_endpoint.js
avranju: That's fine then.

proxy/gateway/nodejs/lib/codec.js
avranju: Yup, saw that. The doc looks great!

proxy/gateway/nodejs/index.js
damonbarry: Good catch. I just pushed another commit that destroys the existing connection (and module) cleanly if the Create control message comes through twice without an intervening Destroy.

PR #311, opened 2017-06-16
Remove filter from message printing

DBojsen:
When using the printer module to debug solutions it doesn't make sence that only messages coming from "Sensor" is printed.

### DIFF
--- a/samples/dotnet_core_module_sample/modules/PrinterModule/DotNetPrinterModule.cs
+++ b/samples/dotnet_core_module_sample/modules/PrinterModule/DotNetPrinterModule.cs
@@ -24,7 +24,17 @@ namespace PrinterModule
 
         public void Receive(Message received_message)
         {
-            Console.WriteLine("Printer Module received message from " + received_message.Properties["source"] + ". Content: " + System.Text.Encoding.UTF8.GetString(received_message.Content, 0, received_message.Content.Length));
+            if (received_message != null)
+            {
+                string messageData = System.Text.Encoding.UTF8.GetString(received_message.Content, 0, received_message.Content.Length);
+                Console.WriteLine("{0}> Printer module received message: {1}", DateTime.Now.ToLocalTime(), messageData);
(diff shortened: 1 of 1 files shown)


samples/dotnet_core_module_sample/modules/PrinterModule/DotNetPrinterModule.cs
```diff
@@ -24,10 +24,7 @@ public void Destroy()
 
         public void Receive(Message received_message)
         {
-            if(received_message.Properties["source"] == "sensor")
-            {
-                Console.WriteLine("Printer Module received message from Sensor. Content: " + System.Text.Encoding.UTF8.GetString(received_message.Content, 0, received_message.Content.Length));
-            }
+            Console.WriteLine("Printer Module received message from " + received_message.Properties["source"] + ". Content: " + System.Text.Encoding.UTF8.GetString(received_message.Content, 0, received_message.Content.Length));
```
aribeironovaes: >Console.WriteLine("Printer Module received message from " + received_message.Properties["source"] + ". Content: " + System.Text.Encoding.UTF8.GetString(received_message.Content, 0, received_message.Content.Length)); [](start = 12, length = 215)

Same comment form other CR. (Thanks for doing this BTW! :))
I don't think "source" property is added automatically. So, on this case you should print the Content separated from the properties.

So, please, do something like this:
if (received_message != null)
{
string messageData = System.Text.Encoding.UTF8.GetString(received_message.Content, 0, received_message.Content.Length);
Console.WriteLine("\t{0}> Received message: {1}", DateTime.Now.ToLocalTime(), messageData);

                    int propCount = 0;
                    foreach (var prop in received_message.Properties)
                    {
                        Console.WriteLine("\t\tProperty[{0}> Key={1} : Value={2}", propCount++, prop.Key, prop.Value);
                    }

aribeironovaes:
:clock1:

PR #313, opened 2017-06-19
To be able to load related dlls

hirosho:
I'm developing HTTP REST API communication module with CPP REST SDK(Casablanca).
I installed Casablanca by vcpkg.
In this case, when I build dynamic library of the module, cpprest_2_9.dll and necessary boost, OpenSSL dll files are installed in the module's ${configuration}(Debug or Release) folder. 
At the time when loader load the dll, loader should load not only module dll but also related dll files but current logic doesn't indicate path of dlls so that loading action fail.
I add logic to get module dll files path and to specify it by SetDllDirectoryA calling.


hirosho:
Thank your for your response.

Yes my PR’s main purpose is for dynamic loading path setting.
REST module is as sample to show this problem. I will move my module to valid position.
Featured Modules’s link is very helpful. There are some friends who want to publish feature modules. I will notice this link to them.

Thanks!

Hiroshi Ota

damonbarry:
No movement on the PR comments, and this probably isn't the right approach anyway. Closing, feel free to reopen if needed...

PR #321, opened 2017-06-26
Allow a device to receive messages before sending telemetry.

darobs:
A device may not ever send telemetry, or send telemetry rarely, but the user might still need to have that device receive messages from the cloud.  The iothub module doesn't know about devices until it is sent a message about the device, so I added the ability to send a registration message to the iothub module.
The message has no telemetry, and does not send a d2c message, but the IoTHub device is created, and can receive messages.

Registration message is recognized by these properties:

>| PropertyName   | Description                                                                  |
>|----------------|------------------------------------------------------------------------------|
>| deviceName     | The device ID registered with IoT Hub                                        |
>| deviceKey      | The device key registered with IoT Hub                                       |
>| deviceFunction | "register"                                                                   |

As an example, Simulated device will send a register message.

### DIFF
--- a/gatewayFunctions.cmake
+++ b/gatewayFunctions.cmake
@@ -113,7 +113,7 @@ function(findAndInstall libraryName version submoduleRootDirectory cmakeRootDire
             #If the library directory doesn't exist, pull submodules
             if(NOT EXISTS "${cmakeRootDirectory}/CMakeLists.txt")
                 execute_process(
-                    COMMAND git submodule update --init ${submoduleRootDirectory}
+                    COMMAND git submodule update --init --recursive ${submoduleRootDirectory}
                     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
                     RESULT_VARIABLE res
                 )
--- a/modules/iothub/devdoc/iothub.md
+++ b/modules/iothub/devdoc/iothub.md
@@ -22,6 +22,16 @@ The IotHub module dynamically creates per-device instances of IoTHubClient. It c
 IoTHubClient. Note that the AMQP and HTTP transports will share one TCP connection for all devices; the MQTT transport will create a new
 TCP connection for each device.
 
+#### Device registration
+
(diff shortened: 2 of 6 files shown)


modules/iothub/devdoc/iothub.md
```diff
@@ -22,6 +22,16 @@ The IotHub module dynamically creates per-device instances of IoTHubClient. It c
 IoTHubClient. Note that the AMQP and HTTP transports will share one TCP connection for all devices; the MQTT transport will create a new
 TCP connection for each device.
 
+#### Device registration
+
+In some circumstances, a device may not want to send telemetry but receive cloud to device messages.  WHen a device wants to connect to the IoTHub but but send a telemetry message, it may send a message with the following properties:
```
avranju: Nit: Minor fixes:

> In some circumstances, a device may not want to send telemetry but receive cloud to device messages.  When a device wants to connect to the IoTHub and wishes to signal that it is expecting to receive cloud-to-device messages, it may send a message with the following properties:

gatewayFunctions.cmake
```diff
@@ -113,7 +113,7 @@ function(findAndInstall libraryName version submoduleRootDirectory cmakeRootDire
             #If the library directory doesn't exist, pull submodules
             if(NOT EXISTS "${cmakeRootDirectory}/CMakeLists.txt")
                 execute_process(
-                    COMMAND git submodule update --init ${submoduleRootDirectory}
+                    COMMAND git submodule update --init --recursive ${submoduleRootDirectory}
```
darobs: iot-sdk-c - it pulls in c-shared code for unit tests.

modules/iothub/devdoc/iothub.md
darobs: That's better - I'll make that change.

gatewayFunctions.cmake
damonbarry: I don't think that's a new dependency. I recall that Andrew did something special in the unit test case.

darobs: We need to build the E2E tests to get the iothub_test library, and the E2E tests are pulling c files from c-shared.  This is probably an issue we need to resolve, so I just posted #323

PR #325, opened 2017-06-28
Add Jenkins script for Win64 VS2017 builds

damonbarry:
We needed a jenkins script for the new Windows x64 VS 2017 gated build we're adding. This build doesn't include remote module support, since that requires libuv, which doesn't yet build with VS2017.


damonbarry:
Nevermind...the driving factor behind a new script was that we needed to avoid building libuv. But I remembered that libuv builds just find when you have both VS 2015 *and* 2017 installed, which the build agent does. So we'll just use the regular windows script for the new VS 2017-based build.

PR #329, opened 2017-06-30
Add Device Twin Desired Properties receiving logic

hirosho:
Hi,
I added Device Twin Desired Properties receiving logic as just a sample.
The identitymap notice to JSON configuration specified deviceId to iothub via broker bus at Create calling.
The iothub receive this message then open an IoT Hub connection for notified deviceId and set Twin Callback. When the Twin Callback is called, the iothub notice received Twin Desired Properties to broker bus.
I think this extension is being developed by SDK team so that If it is already in the mainstream development, I will use it and delete this request.

Regards,

### DIFF
--- a/modules/identitymap/src/identitymap.c
+++ b/modules/identitymap/src/identitymap.c
@@ -491,6 +491,52 @@ static void * IdentityMap_ParseConfigurationFromJson(const char* configuration)
     return result;
 }
 
+static void IdentityMap_Start(MODULE_HANDLE module)
+{
+    IDENTITY_MAP_DATA * idModule = (IDENTITY_MAP_DATA*)module;
+    for (size_t i=0;i<idModule->mappingSize;i++)
+    {
+        IDENTITY_MAP_CONFIG*  macToDevIdArray = (idModule->macToDevIdArray) + i;
+        MESSAGE_CONFIG msgConfig;
+        MAP_HANDLE propertiesMap = Map_Create(NULL);
+        if(propertiesMap == NULL)
+        {
+            LogError("unable to create a Map");
+        }
+        else
+        {
+            if (Map_AddOrUpdate(propertiesMap, "initial-registration", "true") != MAP_OK)
+            {
+                LogError("unable to Map_AddOrUpdate");
+            }
+            else
+            {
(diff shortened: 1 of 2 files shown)


kiranpradeep:
This pull requests lets data acquisition modules receive updates directly from device twin and so appears to address some of the concerns raised here([1](https://github.com/Azure/iot-edge/issues/402)). Why is this not merged?

avranju:
Thanks for the PR. In general, for changes of this nature, I'd suggest that you first create an issue and talk through the design first so we can agree that that's how we want to implement this before you actually do the work. From looking at the code it appears that this is how it works:

1. The identity mapper module generates a message for every entry in its configuration that's intended to be routed to the iothub module with a special property set on the message (called `initial-registration`). This is being done in order to cause the iothub module to open a connection to the cloud on behalf of that device. The implementation constrains this just for the MQTT protocol though (which shouldn't be necessary since twin operations are now supported on AMQP as well). The iothub module handles this message as a special case and responds by creating the connection to the cloud for the device but does not forward the message along.

2. The iothub module registers to be notified for twin desired property updates and for device method calls and responds to that by publishing a message to the broker containing the twin payload or the method call context. For method calls there doesn't appear to be any way for a module receiving that call to provide a response.

For item 1, this functionality has already been implemented in the iothub module as part of PR #321. So we should use that. For 2, I am wondering if we should provide a way for the module to respond to method calls that can then get sent back to iothub. The current design doesn't allow for that.

Other than that the PR has some general issues:

1. Memory leaks in multiple places where a `MAP_HANDLE` is created but not destroyed in certain branches. `MESSAGE_HANDLE` instances get created but don't get destroyed.

2. No tests. All the changes need to have unit tests.

I know it has been a while since you created this PR. Is this something you're willing to address at this point?

hirosho:
Hi Rajasekharan,

This work was done before IoT Hub Device Twin supported only MQTT protocol.
I’ll check current situation and 2nd topic then modify my code including adopt new IoT Edge SDK repository structure.

Thanks!

Hiroshi Ota

damonbarry:
Closing for now. @hirosho Please reopen if you pick this back up. Thanks!

Issue #218, opened 2017-04-19
libuv doesn't build in VS2017


RikVerbrugge:
I am running Windows 10 Creative Update with VS 2015 update 3 and latest versions of cmake and git. But every time I try to build the SDK on Windows 10 I get these errors:

```
   "C:\tmp\azure-iot-gateway-sdk\build\azure_iot_gateway_sdk.sln" (default target) (1) ->
       "C:\tmp\azure-iot-gateway-sdk\build\proxy\modules\native_module_host\native_module_host.vcxproj.metaproj" (defau
       lt target) (29) ->
       "C:\tmp\azure-iot-gateway-sdk\build\proxy\modules\native_module_host\native_module_host.vcxproj" (default target
       ) (46) ->
       (Link target) ->
         LINK : fatal error LNK1104: cannot open file '..\..\..\..\build_libuv\dist\lib\libuv.lib' [C:\tmp\azure-iot-ga
       teway-sdk\build\proxy\modules\native_module_host\native_module_host.vcxproj]

       "C:\tmp\azure-iot-gateway-sdk\build\azure_iot_gateway_sdk.sln" (default target) (1) ->
       "C:\tmp\azure-iot-gateway-sdk\build\core\gateway.vcxproj.metaproj" (default target) (15) ->
       "C:\tmp\azure-iot-gateway-sdk\build\core\gateway.vcxproj" (default target) (41) ->
         LINK : fatal error LNK1104: cannot open file '..\..\build_libuv\dist\lib\libuv.lib' [C:\tmp\azure-iot-gateway-
       sdk\build\core\gateway.vcxproj]
```
I have created several clones on different machines, but every time the same error. The directory where the libuv.lib should stay is also empty. I did a build --rebuild-deps, but got the same error

zafields:
Which version of CMake are you using?

zafields:
Can you specify exactly which versions of git and cmake you are using, so we can reproduce your error?

fernandoBRS:
I'm getting the same issue, using git (2.12.2.windows.1) and cmake (3.8.0).

zafields:
I will setup a VM and try to reproduce your error.

fernandoBRS:
@zafields My python version is 2.7.13. Seems that it's not installing libuv. I needed to copy the libuv.lib to the folder "build_libuv\dist\lib\" at the right moment in the build to make it working.

zafields:
[...]

PR #333, opened 2017-07-05
Lock LibUV to commit w/ VS2017 support

zafields:
Addresses #218, by snapping to latest LibUV release.


zafields:
Migrated changeset to #334

PR #351, opened 2017-07-26
Modify CMake files to install to CMAKE_INSTALL_PREFIX

maartendemunck:
When not installing in /usr (because a different location is required or because root access is not available), the CMAKE_INSTALL_PREFIX variable is not passed to the dependencies, causing the build to fail. This commit passes the CMAKE_INSTALL_PREFIX variable to all child CMake processes. With these changes, a "cmake -DCMAKE_INSTALL_PREFIX=/home/.../local .." does build and install in /home/.../local.

### DIFF
--- a/dependencies-test.cmake
+++ b/dependencies-test.cmake
@@ -6,14 +6,14 @@ include("gatewayFunctions.cmake")
 ###############################################################################
 ############################Find/Install/Build ctest###########################
 ###############################################################################
-findAndInstall(ctest 1.1.0 ${PROJECT_SOURCE_DIR}/deps/ctest ${PROJECT_SOURCE_DIR}/deps/ctest -G "${CMAKE_GENERATOR}")
+findAndInstall(ctest 1.1.0 ${PROJECT_SOURCE_DIR}/deps/ctest ${PROJECT_SOURCE_DIR}/deps/ctest -G "${CMAKE_GENERATOR}" -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX})
 
 ###############################################################################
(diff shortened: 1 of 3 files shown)


darobs:
@maartendemunck Thank you for the contribution!

When we originally envisioned how this might work, we had thought that the developer would build and install their own version of the dependencies separately, and that when the Edge SDK builds, it would find them.  Your change allows the dependencies to be installed during the Edge build, which is useful.

When we do a build where we don't install the submodule dependencies in a global location, we install them into an`install-deps` folder in the repository.  In order to do this, we set a variable, `dependency_install_prefix` which _also_ inserts `-DCMAKE_INSTALL_PREFIX=${dependency_install_prefix}` into the submodule's cmake command. 

So, with this change, we might have two different assignments to `CMAKE_INSTALL_PREFIX` in the cmake call to build the dependencies. It passed the build gate without trouble, so I haven't figured out what the effect will be yet...  I'm looking into that now, and will either approve this or make suggestions for improvement.

I understand this is less than desirable, but I think you can get the same effect right now by setting this on the cmake command line:
```
-DCMAKE_INSTALL_PREFIX=/home/.../local... -Ddependency_install_prefix=/home/.../local...
```

Thanks,

darobs:
@maartendemunck, we've discussed this pull request, and given that is causes the same CMAKE variable to be defined and redefined when building our submodules, I don't think we can merge this as is.

I still like the idea of building the dependencies into the same CMAKE_INSTALL_PREFIX as assigned at the top level, and I've added this request to our backlog.

Unless you want to take on this project, I'm going to close this pull request.

Thank you,

PR #370, opened 2017-08-15
Submodule update to C SDK 2017-08-11 release

darobs:
Update the C-SDK and related submodules to latest release, handle the fallout from the more recent changes.  All tests passed in CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,11 @@ option(use_http "set use_http to ON if http is to be used, set to OFF to not use
 option(use_mqtt "set use_mqtt to ON if mqtt is to be used, set to OFF to not use mqtt" ON)
 option(use_xplat_uuid "use the SDK's platform-independent UUID implementation (default is OFF)" OFF)
 
+if(WIN32)
+    # C-SDK requires this to be set for some functions we use.
+    set(use_etw OFF)
+endif()
+
 option(enable_event_system "Build event system (default is ON)" ON)
 
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -84,6 +84,10 @@ option(use_http "set use_http to ON if http is to be used, set to OFF to not use
 option(use_mqtt "set use_mqtt to ON if mqtt is to be used, set to OFF to not use mqtt" ON)
 option(use_xplat_uuid "use the SDK's platform-independent UUID implementation (default is OFF)" OFF)
 
+if(WIN32)
+    option(use_etw "set use_etw to ON if ETW logging is to be used. Default is OFF" OFF)
```
damonbarry: This part is weird. "If ETW logging is to be used" how? 😮 I'd really like to understand whether we **have** to add this to our root CMakeLists.txt file...I can't imagine we do. Instead, I suspect we can just pass `-Duse_etw=OFF` in the cmake command that builds whichever dependency that requires this.

darobs: It's not in the submodules where we need it - we need it for the `set_platform_files` CMake function, which sets up the C-Shared utility library with the some cached variables.  We call this function to set up those variables used in the `build_test_artifacts` CMake function for unit tests.

[In that function](https://github.com/Azure/azure-c-shared-utility/blob/883b618258097a51bfd226386ca4f04751d11847/configs/azure_c_shared_utilityFunctions.cmake#L482), there's an `if( NOT ${use_etw})` statement which is incorrectly evaluating to false, and so the tests were trying to use the etw_driver rather then the console logger. 

I haven't figured out why `NOT ${use_etw}` would evaluate to false if the option is never set.

darobs: Updated and squashed - "use_etw" is nowjust a variable, set in the top level CMakeLists.txt.

Issue #339, opened 2017-07-12
`activation.type: "launch"` - Out-of-process module exits silently on error and signal conditions


zafields:
Out-of-process modules ALWAYS exit silently. This behavior is fine when a program is signaled by the gateway process or exits normally, but becomes problematic when a failure occurs or a signal is generated externally.

This problem has manifested in two ways:

1. When a process crashes immediately on launch (i.e. unable to load linked library), the gateway will deadlock awaiting the "come alive" ping of the launched process and will not provide any information to the user.
1. When a process is signaled externally or runtime crashes (i.e. bad memory access), the gateway will terminate without providing any status of the external process or indicating the cause for termination.
   - The error message provided, `Error: Time:Wed Jul 12 18:00:56 2017 File:/root/iot-edge/proxy/outprocess/src/module_loaders/outprocess_module.c Func:Outprocess_Destroy Line:955 unable to send destroy control message [0x557584680fa0], continuing with module destroy`, does not help a user understand the reason for exit.

damonbarry:
To clarify, #371 (already merged), fixes #2, so #1 is the only outstanding issue.

Issue #326, opened 2017-06-29
activation.type: launch - proxy_sample gets terminated if we manually kill out of process module


kiranpradeep:
On Ubuntu 14.04 LTS, samples/proxy_sample application will crash if manually kill( kill -9 pid )  the launched process(proxy_sample_remote). This happens only if we had used launch options(**activation.type: launch**).

If we try with activation.type:none and kill the _proxy_sample_remote_ process, the gateway process(proxy_sample) continues with out impact.

zafields:
Are you able to kill without `-9`?

kiranpradeep:
@zafields Yes. I am able to kill with out -9. Copying terminal below

```
username@machine_name:~/kiran/azure-iot-gateway-sdk/build$ ./samples/proxy_sample/proxy_sample ./launch_sample_lin.json 
gateway successfully created from JSON
gateway shall run until ENTER is pressed
Error: Time:Fri Jun 30 10:35:44 2017 File:/home/862537/kiran/azure-iot-gateway-sdk/proxy/outprocess/src/module_loaders/outprocess_module.c Func:Outprocess_Destroy Line:955 unable to send destroy control message [0x7fb89bd6afc0], continuing with module destroy
username@machine_name:~/kiran/azure-iot-gateway-sdk/build$
```

zafields:
I will need to investigate, and I will get back to you.

zafields:
I'm still looking into this... I'm setting up a repro environment. I've also been looking into #320 at the same time to see if there is some sort of relationship.

zafields:
@darobs I've added you to this issue for a morning brainstorming session about the best resolution to this behavior.
@kiranpradeep I have confirmed there is no relationship to #320 and should have this fixed before long.

zafields:
@kiranpradeep The behavior you are witnessing is by design.

When `activation.type: "launch"` is specified, it is assumed the gateway has full control of the life cycle of the module, exactly as if the module was running in the same process. Signaling an out-of-process module with an activation type of launch, is treated the same as if an in-process module received a signal or encountered an error (i.e. bad memory access). When `activation.type: "none"` is specified, then you are assumed to be in control of the life-cycle of the out-of-process module and you can signal, kill, exit the program as you wish.

I believe the real problem is the unhelpful error message you are receiving and I have created an issue #339 to track it.

We would be interested to hear your specific use case and needs to see if we can model your feedback into a future enhancement.

kiranpradeep:
[...]

PR #371, opened 2017-08-17
Log info about the exiting out of process module. Fix for item 2 in #339

kkmsft:


kkmsft:
Applicable to the issue reported in #326

darobs:
Hell @kkmsft - the build gate flagged an error in the unit test compile - it needed a mock of uv_close, so I added one and pushed it to your branch. I will be able to merge this request now.

Issue #378, opened 2017-08-30
retry policy for iothub module


villepalo:
Would it be feasible, to add retry policy  (IOTHUB_CLIENT_RETRY_INTERVAL, ..., OTHUB_CLIENT_RETRY_EXPONENTIAL_BACKOFF_WITH_JITTER) as a new configurable property for iothub module.

darobs:
Hello @villepalo 

Currently, we use the default retry policy available for the transport type.

We have "Expose retry/backoff features in IoT Hub module" in our backlog, but I don't have an estimated time for completion.

villepalo:
Hi @darobs 

I already started to implement this kind of feature. It is 'must have' feature for me.

PR #382, opened 2017-09-07
Configrable RetryPolicy for IotHub -module

villepalo:
* Added RetryPolicy configuration json field
* devdoc updates
* Updated old ones and added some new unit tests
* Updated configuration json files of all samples using IotHub module

Issue refence: #378


darobs:
Only one comment so far - instead of requiring the retry policy in the JSON, we'd prefer a default if it's not set.  That would better maintain backward compatibility.

(The default should be "EXPONENTIAL_BACKOFF_WITH_JITTER").

villepalo:
I thought about it long and hard and then made the wrong decision :)

darobs:
Thank you @villepalo, I'll run this through our build gate and let you know if any trouble comes up.

PR #386, opened 2017-09-12
Update to latest master in iothub sdk, pull in latest submodules to do so.

darobs:
use iothub_client shared library's new name.
fix strftime in logger

The C SDK moved parson, need to fix our build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,10 +266,10 @@ endfunction(copy_module_host_dll)
 
 function(copy_iothub_client_dll whatIsBuilding whatIsBuildingLocation)
   if(WIN32)
-    if(EXISTS "${azure_iot_sdks_DIR}/../bin/iothub_client.dll")
+    if(EXISTS "${azure_iot_sdks_DIR}/../bin/iothub_client_dll.dll")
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -266,10 +266,10 @@ endfunction(copy_module_host_dll)
 
 function(copy_iothub_client_dll whatIsBuilding whatIsBuildingLocation)
   if(WIN32)
-    if(EXISTS "${azure_iot_sdks_DIR}/../bin/iothub_client.dll")
+    if(EXISTS "${azure_iot_sdks_DIR}/../bin/iothub_client_dll.dll")
```
damonbarry: I thought they changed the target to `iothub_client_dll`, but still kept the binary as `iothub_client.dll`?

darobs: They couldn't. Creating a Windows DLL actually creates 2 files, <name>.dll and <name>.lib (3 if you count the PDB), which would conflict with the static library <name>.lib. To avoid the filename conflict, the C SDK doesn't rename the shared library.  It does on Linux, which is perfectly happy with its independent .so and .a files.

damonbarry: Ah, right. I remember now that we talked about this.

Issue #391, opened 2017-09-15
'Message Delivered'-notification from IotHub


villepalo:
Hi,

We need a 'message delivered' notification from IotHub when message is successfully delivered. 

Our 'buffer'-module stores messages to database before sending them to IotHub and removes them after message is successfully delivered. Hence we need some kind of notification from IotHub about delivered messages. 

Are you planning to implement or interested in having this kind of mechanism (message delivered notification) in the IotHub module?

myagley:
Closing as the PR was merged.

PR #392, opened 2017-09-19
Iothub delivery notification

villepalo:
Here is what I was talking about in #391 ('Message Delivered'-notification from IotHub). This feature is mandatory for us, so I just implemented it. If you are interested in having this kind of feature, here it is. If this is not useful to you, feel free to say no.

### DIFF
--- a/modules/iothub/src/iothub.c
+++ b/modules/iothub/src/iothub.c
@@ -709,58 +709,6 @@ static void SendEventAsync_receiveMessageConfirmation(IOTHUB_CLIENT_CONFIRMATION
         return;
     }
 
-    char* deliveryStatus;
-    switch (result)
-    {
-        case IOTHUB_CLIENT_CONFIRMATION_OK:
-        {
-            if (mallocAndStrcpy_s(&deliveryStatus, "OK") != 0)
-            {
-                LogError("Cannot copy deliveryStatus code");
-                return;
-            }
-            break;
-        }
-        case IOTHUB_CLIENT_CONFIRMATION_BECAUSE_DESTROY:
-        {
-            /*Codes_SRS_IOTHUBMODULE_99_007: [ If "iotHubMessageId" is set and message is not delivered successfully 'message delivered' notification is sent with "deliveryStatus" property set "DESTROY", "TIMEOUT" or "ERROR" ]*/
-            LogError("Message was not delivered due to IOTHUB_CLIENT_CONFIRMATION_BECAUSE_DESTROY");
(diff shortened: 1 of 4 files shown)


modules/iothub/src/iothub.c
```diff
@@ -759,8 +913,40 @@ static void IotHub_Receive(MODULE_HANDLE moduleHandle, MESSAGE_HANDLE messageHan
...
+
+                                    if (mallocAndStrcpy_s(&(userContextCallback->iotHubMessageId), iotHubMessageId) != 0)
+                                    {
+                                        /*Codes_SRS_IOTHUBMODULE_99_008: [ If memory allocation fail when handling "iotHubMessageId" property, `IoTHubClient_SendEventAsync` returns without sending the message ]*/
+                                        LogError("Failed to allocate/copy iotHubMessageId");
+                                        free(userContextCallback);
+                                        IoTHubMessage_Destroy(iotHubMessage);
+                                        ConstMap_Destroy(properties);
+                                        return;
+                                    }
```
darobs: Do we want to return here (and in the section above) or make a best effort at sending the message to IoTHub?

modules/iothub/src/iothub.c
```diff
@@ -686,6 +701,145 @@ static IOTHUB_MESSAGE_HANDLE IoTHubMessage_CreateFromGWMessage(MESSAGE_HANDLE me
     return result;
 }
 
+static void SendEventAsync_receiveMessageConfirmation(IOTHUB_CLIENT_CONFIRMATION_RESULT result, void* userContextCallback)
+{
+    if (!userContextCallback)
+    {
+        LogError("Context was null");
+        return;
```
darobs: >return [](start = 8, length = 6)

The coding standards for our C code is to have one return call.

modules/iothub/src/iothub.c
villepalo: Good question. If malloc fails how likely it is that sending message will success? If user needs a notification about delivery, and doesn't get one, wouldn't that be a huge problem too?

modules/iothub/src/iothub.c
villepalo: fixed

modules/iothub/src/iothub.c
darobs: The most likely way is if the "iotHubMessageId" string gets trashed in memory and the malloc_and_Strcpy_s fails, but even then it's unlikely that the IoTHubClient_SendEventAsync will succeed as well.  

So it's pretty unlikely it worth trying to send the message. I'm OK with leaving it alone, just more curious if you had a different answer..

Issue #394, opened 2017-09-24
Why source property must be "mapping" in IotHub module


villepalo:
Is there any good reason, why the "source" property must be set to "mapping". I mean, if I do not need IdentityMapper module I still have to set this property as "mapping". Shouldn't links array in configuration file take care of this kind of routing? I think setting deviceName and deviceKey properties should be enough. Or am I missing something here?

darobs:
It's a leftover bit of code from when we didn't have routing. There's no need for it now. 

...although, it's also a check in case you set up a route that sends messages to the iothub module you didn't actually want to sent to IoTHub. I like the requirement that a message be flagged for transmission to the IoTHub, but you could argue that a message which has device ID and device key is already flagged for transmission to the IoTHub.

We do have a PBI for this.  The code & test of it is pretty easy, but we'd have to confirm that we are not breaking anyone by removing the check.

kiranpradeep:
@darobs Does the identity mapper module needs the source property(mapping) now? A devdoc([1](https://github.com/Azure/iot-edge/blob/master/modules/identitymap/devdoc/identity_map.md#identity-map-module-requirements)) comment mentions it was meant for a module to reject it's own messages.

darobs:
@kiranpradeep The pull request is there, but we haven't merged it yet - we've been pretty heads down on the V2 Edge features.  This is a problem we had before routes were introduced. With the routes in place, this would only be a problem if you routed iothub module messages directly back to the iothub module, which would be an odd thing to do.  I was intending on accepting #395.

kiranpradeep:
@darobs The pull request doesn't touch identity mapper module. It only changes iothub module. For completion, shouldn't it change identity mapper too?

villepalo:
@kiranpradeep Good point! I'm not sure though. Some one could use the source property for other purposes as well. It might be, that someone has created a module which depends on identitymapper-module and uses this property. I don't know, maybe @darobs knows better :)

BTW, shouldn't this question/thread be in review comments, not here?

PR #395, opened 2017-09-26
Iothub remove mandatory source

villepalo:
Removed mandatory source = mapping property. See #394

### DIFF
--- a/modules/iothub/devdoc/iothub.md
+++ b/modules/iothub/devdoc/iothub.md
@@ -12,7 +12,6 @@ send a message with the following properties:
 
 >| PropertyName | Description                                                                  |
 >|--------------|------------------------------------------------------------------------------|
->| source       | The module only processes messages that have source set to "mapping".        |
 >| deviceName   | The device ID registered with IoT Hub                                        |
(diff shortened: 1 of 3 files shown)


kiranpradeep:
The pull request doesn't touch identity mapper module. It only changes iothub module. For completion, shouldn't it change identity mapper too?

villepalo:
It could be that some module uses identityMapper source property for other purposes, it is not necessary for iothub module. So I think idenitymapper module should not be modified in this pull-request.

damonbarry:
For better or worse, source is not part of the message contract for the iothub module. Wouldn't it be a breaking change to remove it? If so, we probably shouldn't take this change.

darobs:
Thank you so much for your contribution. Unfortunately, I'm agreeing with @damonbarry on this.  We're not going to merge this as is out of concern for breaking existing applications, but I think the problem as described does have merit. I did add to our backlog a suggestion:

> Add an optional configuration (something like "AcceptModule": "<module name>" ) that allows IoTHub to accept a message from source=="<module name>"  default will be to accept source==mapping so that changes will not break existing applications.

PR #486, opened 2018-01-23
Add avoid message lost of nanomsg features by thread messaging mechanism

hirosho:
Current broker runtime uses nanomsg as messaging mechanism but nanomsg causes many messages lost because it uses pub/sub mode.
Worst case, 40-50% messages have been lost.
I add thread messaging mechanism for communication between two modules by add-link specification in JSON configuration. 
When user specifies following in JSON configuration...
"links":[
 { "source":"SensorTag", "sink":"mapping", "message-type":"thread-message" }
And both source and sink are loaded as in-process then broker use not nanomsg but thread messaging mechanism.

### DIFF
--- a/v1/core/inc/broker.h
+++ b/v1/core/inc/broker.h
@@ -30,6 +30,15 @@ extern "C"
 #include <stddef.h>
 #endif
 
+#define BROKER_LINK_MESSAGE_TYPE_VALUES \
+    BROKER_LINK_MESSAGE_TYPE_DEFAULT, \
+    BROKER_LINK_MESSAGE_TYPE_THREAD
+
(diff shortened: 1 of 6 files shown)


damonbarry:
@hirosho You're right, nanomsg pub/sub does not provide delivery guarantees--it is best-effort. You can minimize the risk by architecting your gateway in such a way that the modules don't send messages faster than your networking stack can process them. We've only seen dropped messages in our stress testing when we've really hammered the system as hard as we can. A thoughtfully designed pipeline and hardware to match the load should mitigate the risk.

This is an interesting addition to the v1 gateway, but probably not a feature we want to bring in and support at this time. And if we were to merge it, you'd need to add a lot of tests first. I'll close this PR, but we can continue to discuss the design if you'd like. Thanks!

PR #552, opened 2018-03-22
V2 Samples - Clarifying functionality

dariuszparys:
The deployed temperature sensor has fewer functionality then the source code provided by this sample. I modified the README to clarify which functions are available in the deployed version and which are new.

Please check @varunpuranik @darobs @damonbarry if this makes sense.


varunpuranik:
Changes look good.. thanks.

Issue #280, opened 2017-05-23
build.sh clones libuv for each build


kiranpradeep:
Every run of `build.sh` deletes and clones libuv. Is this intentional ?

damonbarry:
We'd be happy to have something more refined, but the "brute force" approach seemed acceptable as a first step.

myagley:
Contributions welcome. Closing for now.

kiranpradeep:
Can I try a pull request with libuv as a submodule in v1/deps folder? Likes to know the preferred way on this.

darobs:
The cloning and building of libuv was intentional, because libuv is only needed for the out of process modules, and not everyone wants that. It also seems to be the pattern  However, building _every time_ is an annoyance we didn't find a great solution for. 

Yes please, pull requests are welcome.

damonbarry:
I have a pull request nearly ready for this, but I’ve only tested on Mac. I can create the PR so people can look at it, then I can add the Windows/Linux support when I get time (or someone else can add it).

PR #568, opened 2018-04-03
Build libuv using the same CMake machinery as other dependencies

damonbarry:
These changes are incomplete--they work on macOS but I haven't added support yet for Linux or Windows (unfortunately the build instructions for libuv are a little different on each platform). But a related conversation came up in #280, so I'll post this now for people to look at.


damonbarry:
I've updated the build for Windows and Linux too. I've confirmed that all platforms build successfully with their build scripts (build.sh or build.cmd), and that `proxy_sample` works.

damonbarry:
I pushed a few more commits:
- I turned native remote module support OFF by default in CMake, to be consistent with other features. That means `--disable-native-remote-modules` is no longer valid in the build.(sh|cmd) build scripts. Instead, there's a new flag: `--enable-native-remote-modules`.
- If you already have libuv installed (from an installer or package manager, or because you built it from source and did `make install` to the system default location, e.g. /usr/local), then we'll use that instead of trying to build/install it ourselves. This is consistent with how we handle our other dependencies.

**NOTE** on macOS I had to upgrade XCode developer command line tools in order to get past a weird CMake error. I did `xcode-select --install` from the terminal, and it launched a UI which upgraded the tools to being them in line with the latest XCode (which was recently updated on my Mac).

v1/samples/dotnet_core_dynamic_add_module/README.md
```diff
@@ -24,7 +24,7 @@ Building the sample (Linux)
 ---------------------------
 The sample code and dependancies gets build by running:
 ```
-tools\build.sh --enable-dotnet-core-binding --disable-native-remote-modules
+tools\build.sh --enable-dotnet-core-binding
```
darobs: Drop the `--enable-dotnet-core-binding ` here?

damonbarry: This sample (`dotnet_core_dynamic_add_module`) is specifically about the .NET Core binding, so I think it's correct as-is.

Issue #567, opened 2018-04-02
[V1] Upgrade request: dotnetcore modules to based on netcore2.0


kiranpradeep:
In the 6 csproj files in `v1/bindings/dotnetcore` and `v1/samples` folders, I edited the `<TargetFramework>` tag and was able to build dotnetcore modules based on 2.0. I also removed both `<RuntimeFrameworkVersion>` and `<PackageTargetFallback>` tags present in csproj files. 

I am _very_ new to dotnet and likes to know if this Is this the correct way to upgrade to netcore2.0? If yes, I will put a pull request. I will test on both Ubuntu 16.04 and macOS 10.12.3.

darobs:
Yes, please do the pull request.

damonbarry:
@kiranpradeep closed the PR, so I'll close this as well.

PR #583, opened 2018-04-23
netcore 2.0

kiranpradeep:
# Reference/Link to the issue solved with this PR (if any)
#567

# Description of the problem
.NET Core 2.0 changes

# Description of the solution
1. Using dotnetcore 2.0.7 runtime
2. Samples and tests moved to netstandard 2.0/netcore 2.0
3. Nuget Microsoft.Azure.Devices.Gateway left unchanged on netstandard 1.3

## Notes on Unit Tests 
1. Tested only on Ubuntu 16.04 LTS and macOS 10.12.3 (flags: --run-unittests --run-e2e-tests). All unit tests passed. All e2e tests except gateway_e2e passed (no connection string in environment).

## Not sure
1. Not tested on Windows
2. `_NSGetExecutablePath` failure not checked as was the case for nearby `readlink` with comment
3. macOS platform preprocessor (APPLE) in source code as the file already had the same preprocessor check
4. csproj versions changed to 1.0.6


damonbarry:
Upgrading our stuff to .NET Core 2.0 is great! But as I understand it, there's no real benefit to upgrading the _.NET Standard_ bits to 2.0 unless your library is calling 2.0 APIs. A .NET Standard 1.3 library will run perfectly well in a .NET Core 2.0 environment. So I'm wondering if we can keep our .NET Standard libraries on 1.3.

I'm also wondering whether upgrade will break people. We may need to be able to build either 1.x or 2.0 based on a switch.

PR #606, opened 2018-05-17
Added Message ContentEncoding and ContentType

sebader:
Required properties to enable routing on those messages with body content

# Description of the problem
Constructed message does not contain ContentType and ContentEncoding. Hence conditional message routing based on message body content is not possible in IoT (Edge) Hub

# Description of the solution
Just added the two missing properties to the message header


dariuszparys:
@varunpuranik PR is good for me, there is just the caveat that this repo isn't building `microsoft/azureiotedge-simulated-temperature-sensor:1.0-preview` where the actual problem happens when running the docs tutorials. Can we make this repo the building base for the deployed container?

/cc @damonbarry @chipalost

PR #666, opened 2019-03-13
Add YAML builds for Node.js lib, traceability tool

damonbarry:
This change adds YAML files for two new builds in Azure DevOps:
- Build Node.js as a SO/DLL for Windows, Ubuntu, and Debian, and make them available as artifacts. Our IoT Edge v1 builds will consume them in order to build/test Node.js bindings. This is a manual build.
- Build the traceability tool from the Azure IoT C SDK's repo (https://github.com/Azure/azure-iot-sdk-c/tree/master/tools/traceabilitytool) and publish it as an artifact. Our IoT Edge v1 builds will use the tool to verify requirements vs. code/tests. This is a manual build.


damonbarry:
[Build Status](https://iotedge.visualstudio.com/iotedge/_build/results?buildId=7513)

[Build Status](https://iotedge.visualstudio.com/iotedge/_build/results?buildId=7515)

libnode.yml
```diff
@@ -0,0 +1,71 @@
...
+  displayName: Ubuntu 16.04
+  pool:
+    vmImage: ubuntu-16.04
+  steps:
+  - script: |
+      v1/tools/build_nodejs.sh
+      # Convert thin archives to normal ones
+      libs=`find v1/build_nodejs/dist/lib/ -name '*.a'`
+      for lib in $libs; do
+        ar -t $lib | xargs ar -rs $lib.tmp && mv $lib.tmp $lib
```
avranju: Curious, what's this command doing?

damonbarry: Ah, right. When you use `./configure --shared` to set up the Node.js build, it builds the static libraries as thin archives: they still reference the individual object files (which were not copied into `dist/lib`). That's fine when you turn around and consume those libraries in another build on the same machine, but I needed to publish them as artifacts that other builds (other machines) can consume. I used these commands to make the libraries self-contained. The first `ar` command lists all the object files referenced by a given {lib}.a, and the second command inserts them into the library and updates the index.

avranju: Interesting. So just having the `.so` file is not sufficient? Didn't know that!

PR #667, opened 2019-03-19
Add definition for CI build

damonbarry:
This change adds the Azure Pipelines build definition that builds IoT Edge v1 and runs all tests, **including end-to-end**, when changes are pushed to master. The build is **not** triggered by pull requests. The pipeline that references this definition lives in the product team's internal DevOps project to protect against test secrets leaking into public-facing logs.

The build runs on Debian 8, Ubuntu 16.04, and Windows Server 2016. The YAML file has a lot in common with iotedge.yml, so it would be good to use job-level templates to reduce duplication. However I don't anticipate this repo will see a lot of churn so I chose not to invest in templates.

Note: I had to upgrade our use of .NET Core to a newer version of 1.x. The one we were referencing isn't available on build agents.


iotedge-ci.yml
```diff
@@ -0,0 +1,241 @@
...
+        --enable-nodejs-remote-modules
+    displayName: Run the build
+  - task: CopyFiles@2
+    displayName: Collect build logs
+    condition: always()
+    inputs:
+      SourceFolder: v1/deps
+      Contents: "**/build/?(output.txt|error.txt)"
+      TargetFolder: $(Build.ArtifactStagingDirectory)
+  - task: PublishBuildArtifacts@1
```
kkdawkins: Would it ever be useful to publish more than logs? Perhaps publishing the built bits for some sort of longer running integration test on Azure VMs?

v1/bindings/dotnetcore/dotnet-core-binding/Microsoft.Azure.Devices.Gateway.Tests/Microsoft.Azure.Devices.Gateway.Tests.csproj
```diff
@@ -7,7 +7,7 @@
     <PackageId>Microsoft.Azure.Devices.Gateway.Tests</PackageId>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <PackageTargetFallback>$(PackageTargetFallback);dotnet5.4;portable-net451+win8</PackageTargetFallback>
-    <RuntimeFrameworkVersion>1.1.1</RuntimeFrameworkVersion>
+    <RuntimeFrameworkVersion>1.1.11</RuntimeFrameworkVersion>
```
kkdawkins: Any reason you picked 1.1.11? They released one more point on the 1.1.x line -> 1.1.12. There's also Net core 2.x out now since Net Core 3.x is coming next month.

kkdawkins: Looks like there is a NuGet tampering fix that affects linux that is fixed in 1.1.12 https://github.com/dotnet/core/blob/master/release-notes/1.1/1.1.12/1.1.12.md

iotedge-ci.yml
damonbarry: Realistically, I don't think we'll be adding any more build/test automation than what's here.

v1/bindings/dotnetcore/dotnet-core-binding/Microsoft.Azure.Devices.Gateway.Tests/Microsoft.Azure.Devices.Gateway.Tests.csproj
damonbarry: I only upgraded to get onto the current build servers. 1.1.12 is the natural choice, but IIRC it wasn't available in the default location on the agent--1.1.11 was. I could easily get 1.1.12 by using the installer task, but it gets installed at a different location, and our code/build isn't set up to find it there.

